Issue #278, opened 2026-06-10
i18n: refresh 2,289 genuinely-stale translations (pre-existing, not #272-induced)


pjt222:
## Background

During the #272 normalization (PR #276), the i18n freshness pass distinguished two kinds of staleness:

- **Normalization-induced false staleness** — 719 main-synced translations whose only "change" was the formatting normalization itself. These were bumped in `95e380cc` (honest, no content gap).
- **Pre-existing genuine staleness** — translations whose English sources changed in content *before* the normalization. These were deliberately left stale so the status files stay honest (see `feedback_status_yml_lies` lesson: bumping them would mask real content gaps).

This issue tracks the second group: a real translation-refresh effort.

## Scope (verified 2026-06-10 against honest status @ `95e380cc` + committed compressed-locale status)

| Locale | Stale | Breakdown |
|---|---|---|
| caveman | 295 | skills only |
| caveman-lite | 295 | skills only |
| caveman-ultra | 295 | skills only |
| wenyan | 295 | skills only |
| wenyan-lite | 295 | skills only |
| wenyan-ultra | 292 | skills only |
| de | 135 | 133 skills + 1 team + 1 guide |
| es | 138 | 136 skills + 1 team + 1 guide |
| ja | 121 | 119 skills + 1 team + 1 guide |
| zh-CN | 128 | 126 skills + 1 team + 1 guide |
| **Total** | **2,289** | |

Compressed locales sit broadly at `source_commit: 82c77053`; prose locales largely at `6f65f316`.

## Also covered

The 6 caveman/wenyan copies of `create-skill` were left **without** the styleguide cross-reference added in `496be0a2` (the 4 prose-locale copies were refreshed in `95e380cc`). Fold the cross-ref in when refreshing those 6.

## Approach notes

- Refresh = re-translate changed sections + bump `source_commit`, per `i18n/README.md` workflow
- Translated SKILL.md files must stay under the i18n line budget (English + frontmatter overhead, see #266/#271)
- Skills translations use Shape A frontmatter (locale fields under `metadata`)
- Batch by locale; `npm run validate:translations` + `npm run translation:status` after each batch

## Non-goals

- Not caused by and not blocking #272 (AC#4 tracked there)
- No source_commit bumps without actual content refresh

Refs #272, #258.

pjt222:
[...]

PR #348, opened 2026-07-16
docs(repo-protection): fold in session-verified ruleset insights

pjt222:
Updates the guide + `harden-github-repo-security` skill with operational facts established by **actually applying** the protection to this repo this session (dogfood → docs).

## Changes
- **Bypass is whole-ruleset, not per-rule.** A bypass actor skips the *entire* ruleset (incl. `deletion`/`non_fast_forward`), so a single-ruleset bypass re-opens force-push/delete. Added the **two-stacked-rulesets** pattern (ref-protection with no bypass + required-check with the bypass) to the guide and harden Step 3b, plus a troubleshooting row.
- **Deploy-key wiring.** The matrix named the deploy-key option but not the wiring: `checkout` `ssh-key:` → `git-auto-commit-action` pushes over SSH as a `DeployKey` bypass actor. Added to the guide + harden Step 3a. Documented the quirk that a `DeployKey` bypass stores `actor_id: null` (matches any write deploy key).
- **Corrected stale personal-repo framing.** The guide claimed the bypass mechanism is often *unsatisfiable* on personal repos (web picker missing → move to an org). Verified false this session: the **REST API** accepts `Integration`/`DeployKey`/`User`/`RepositoryRole` bypass actors on a free personal public repo. Fixed the caveat, 2 troubleshooting rows, and the harden Step 3 On-failure — the org move is now framed as a last resort.

## Notes
- Grounded in the real deploy-key bypass applied to `pjt222/agent-almanac` (ruleset split into `protect-main-refs` + `require-line-endings`).
- Line counts: guide 286, harden skill 381 (< 500). content-style 0 blocking, line-endings LF, validate:integrity unaffected (no frontmatter/cross-ref/registry change).
- The de/zh-CN/ja/es stubs for these two files go stale (untranslated), consistent with the existing untranslated corpus (tracked under #278); no re-scaffold.

### DIFF
--- a/guides/protecting-github-repositories.md
+++ b/guides/protecting-github-repositories.md
@@ -169,7 +169,7 @@ So to let trusted automation through, you must give it a *real* bypass-eligible 
 | Admin/fine-grained **PAT** + `Repository admin` role in bypass | **No** | 5 | Worst. Role-wide hole — exempts *all* admin actions including your own interactive pushes, not just automation. A leaked PAT = admin bypass-write. |
 | Add the default **`GITHUB_TOKEN`** to the bypass list | — | 6 | **Impossible.** Not a selectable actor by design; the push `403`s. |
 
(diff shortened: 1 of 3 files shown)


pjt222:
Added a third file to this PR: **`assess-github-repo-security`** now notes that a branch can be protected by **multiple stacked rulesets** (GitHub aggregates them, most-restrictive-wins) — the audit must inspect and aggregate *every* active ruleset, not just the first, since the recommended split puts ref-protection and the required check in separate rulesets (Step 3 note + Validation item).

Issue #357, opened 2026-07-17
ci: update-readmes file_pattern is a hand-maintained allowlist that can silently drift from generate-readmes.js output


pjt222:
## Gap

`.github/workflows/update-readmes.yml` commits generated files via `stefanzweifel/git-auto-commit-action` with an explicit `file_pattern` allowlist. That allowlist is maintained **by hand** and is **not** derived from the set of files `scripts/generate-readmes.js` actually writes.

Failure mode: register a new generated file/section in `generate-readmes.js` but forget to add its path to `file_pattern`. On push-to-main the runner regenerates the file correctly, but git-auto-commit-action stages only `file_pattern` paths, so the change is discarded and the file re-drifts on every subsequent registry push — never converging, with no error surfaced.

This just bit #353: `guides/quick-reference.md` was registered in the generator but was absent from `file_pattern` (fixed in the #353 PR by adding it, but the *class* of gap remains for the next generated file).

## Options

- **Single source of truth**: have `generate-readmes.js --list` print the files it manages, and feed that list to `git add` in CI instead of a hardcoded `file_pattern`.
- **CI assertion**: a `validate-integrity` / CI check that every file registered in `generate-readmes.js` (the `processFile` / `writeGeneratedFile` call sites) appears in `update-readmes.yml`'s `file_pattern`.
- **PR gate**: run `npm run check-readmes` as a PR check so a stale generated file is caught at review time regardless of the auto-commit allowlist. (Note: currently *no* generated file is PR-gated; the repo relies on the post-merge auto-commit by design, so this would be a philosophy change.)

## Context

Surfaced by the adversarial review of #353. Low urgency (each occurrence is a one-line fix once noticed), but silent non-convergence is exactly the drift the AUTO-generation machinery exists to prevent.

Issue #354, opened 2026-07-16
quality: team skills' allowed-tools omit Agent/SendMessage/Task* named in their procedures


pjt222:
## Gap

After #337, the activation procedures in `skills/create-team/SKILL.md` (Step 10) and `skills/test-team-coordination/SKILL.md` (Step 4.3) instruct spawning members via the **Agent tool**, coordinating with **SendMessage**, and (create-team) creating tasks via **TaskCreate** — but both declare `allowed-tools: Read Write Edit Bash Grep Glob`, which omit `Agent`, `SendMessage`, and `TaskCreate`.

This is **pre-existing** (the prior TeamCreate-era text was equally absent from `allowed-tools`), surfaced by the #352 review. `skills/unleash-the-agents/SKILL.md` is the counter-example done right — it lists `Agent ToolSearch TeamCreate TaskCreate TaskUpdate TaskList SendMessage`.

## Acceptance Criteria

- [ ] Reconcile `allowed-tools` with the procedure for `create-team` and `test-team-coordination` (add `Agent`, `SendMessage`, and `TaskCreate` where the steps use them).
- [ ] Sweep other skills whose procedures name orchestration tools (Agent/SendMessage/Task*/ToolSearch) for the same drift.
- [ ] Consider a lightweight `validate-integrity` check: tool names in a SKILL.md procedure body should appear in its `allowed-tools`.

## Context

Found via the #352 adversarial review (Finding 4).

PR #369, opened 2026-07-19
ci: warn-only A9 invocation-phrase allowed-tools check — closes #356

### LINKED ISSUE #356: quality: consider a narrow invocation-phrase check that a skill's allowed-tools covers the orchestration tools its procedure invokes (#354 follow-up)
## Gap

#354 reconciled `allowed-tools` for `create-team` and `test-team-coordination` and swept the other 366 skills by hand (no further genuine drift found). Its AC bullet 3 asked to "consider a lightweight validate-integrity check" that tool names in a SKILL.md procedure appear in `allowed-tools`.

A **naive substring check was deliberately not added** in #354: 4 of 6 raw token hits in the sweep were non-invocations — deprecation prose ("`TeamCreate` is gated…"), the A2A `tasks/get` JSON-RPC method, a `subagent_type:` CONFIG schema field, and the Workflow `agent()` in-script primitive. Even the two fixed skills legitimately retain `TeamCreate`/`ToolSearch` deprecation prose a token check would wrongly demand be declared. ~67% false-positive rate → net-negative.

## Possible narrower check (unproven)

Key on the repo's actual authoring convention rather than bare tokens — match invocation phrasings such as `via the Agent tool (`` `subagent_type` ``)`, "coordinate … with `SendMessage`", "create tasks via `TaskCreate`" — and assert the named tool is in `allowed-tools`. This would likely catch the #354 drift class with low FP without parsing intent, but it is untested against the corpus and phrasing-brittle.

## Acceptance

- [ ] Prototype the phrase heuristic against all skills; measure FP / FN.
- [ ] If FP is acceptably low, add as a warn-only `validate-integrity` check (e.g. A8).
- [ ] Otherwise close as "not worth it" with the measurement recorded.

## Context

#354 follow-up. Filed per the "issue for every gap" rule (adversarial-review suggestion) rather than silently dismissing AC bullet 3.

pjt222:
Implements #356's acceptance criteria: prototype the narrow invocation-phrase heuristic, measure FP/FN, and add it as a warn-only integrity check.

## Measurement (2026-07-19, all 368 skills)

- **Pattern**: invocation verb (`via|using|with|through|call*|invoke*|spawn*`) + optional `the` + backticked known orchestration tool name (`Agent|SendMessage|TaskCreate|TaskUpdate|TaskGet|TaskList|TeamCreate|TeamDelete|Workflow`).
- **Hits**: 3 files / 11 lines — `create-team`, `test-team-coordination`, `unleash-the-agents`. All invoked tools are declared in the respective `allowed-tools` → **0 violations, 0 false positives** on the current corpus.
- **Anchoring works**: `memex`'s "via the MCP tool" (not a known tool name) and `test-team-coordination`'s parenthetical TeamCreate deprecation note (no invocation verb) are correctly excluded.
- **FN check**: both historical #354 cases (`create-team`/`test-team-coordination` pre-PR #358) would have been flagged — verified by negative test (removing `TaskCreate` from `create-team` produces the WARN). Un-backticked phrasing (e.g. `unleash-the-agents` line 118 "with the Agent tool") evades the line-level pattern — accepted FN class; the same file is still caught via its backticked lines. This FN class is why the check is warn-only rather than blocking.

This clears the issue's own "FP acceptably low" bar, so the check ships as **A9 (warn-only)** in `scripts/validate-integrity.sh`. (The issue suggested the "A8" slot — taken by #357 in the meantime.)

## Implementation notes

- Shortlist-first: one repo-wide `grep -rlE`, then per-hit-file extraction — negligible runtime cost.
- Portability gotcha caught in testing: an escaped backtick (`` \` ``) is a **buffer-start anchor** in GNU grep (CI) but a literal in ugrep (this WSL's interactive default) — the pattern matched locally and would have silently never matched in CI. The backtick now lives in a single-quoted variable, verified against `/usr/bin/grep`.
- Warn semantics follow the established `warn_count` pattern (B5/B7-B9); `failed` untouched.

## Verification

- Positive: full A-section run green, `OK: no invocation-phrase/allowed-tools drift`
- Negative 1: `TaskCreate` dropped from `create-team` → WARN
- Negative 2: invocation line added to a skill with no `allowed-tools` frontmatter → WARN

Closes #356.

### DIFF
--- a/scripts/validate-integrity.sh
+++ b/scripts/validate-integrity.sh
@@ -222,30 +222,39 @@ fi
 # A skill whose procedure invokes an orchestration tool by name ("via the
 # `Agent` tool", "coordinate ... with `SendMessage`") should declare that tool
 # in its frontmatter allowed-tools — the #354 drift class. The heuristic is
-# deliberately narrow: an invocation verb, optional "the", then a backticked
-# KNOWN orchestration tool name. Measured 2026-07-19 over all 368 skills:
-# 3 files / 11 lines matched, 0 false positives, 0 violations; both historical
-# #354 cases would have been flagged. The verb anchor excludes bare mentions
-# (deprecation notes, tool lists); un-backticked phrasing evades the pattern —
-# accepted false-negative class, which is why this stays warn-only (see #356).
+# deliberately narrow: an invocation verb (lower- or sentence-case), optional
+# "the", then a backticked KNOWN orchestration tool name. Measured 2026-07-19
+# over all 368 skills: 3 files / 12 lines matched, 0 false positives,
+# 0 violations; both historical #354 cases would have been flagged. The verb
+# anchor excludes bare mentions (deprecation notes, tool lists). Known
+# false-negative classes, accepted because this is warn-only (see #356):
+# un-backticked tool names ("with the Agent tool"), markdown emphasis breaking
+# adjacency ("**via** `Agent`"), and verbs outside the list.
+# allowed-tools is parsed from the frontmatter block only (first ---...---),
+# in both inline (space-separated) and YAML block-list form.
 echo "--- A9: Invocation-phrase allowed-tools coverage (warn-only) ---"
 a9_tools='Agent|SendMessage|TaskCreate|TaskUpdate|TaskGet|TaskList|TeamCreate|TeamDelete|Workflow'
-a9_verbs='via|using|with|through|call|calls|calling|invoke|invokes|invoked|spawn|spawns|spawned|spawning'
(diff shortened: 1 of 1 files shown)


pjt222:
Adversarial review (advocatus-diaboli; Copilot quota-blocked): **APPROVE-WITH-NITS**, 5 findings, none blocking. All folded in 30c02968: frontmatter-anchored allowed-tools parse supporting YAML block-list form (findings 1+2 — kills the latent block-form false positive), sentence-case + use/uses/used verbs (finding 3 — "Use the `Agent` tool" now caught; re-measured 3 files / 12 lines / 0 FP), corrected line count (finding 4), dropped redundant -r (finding 5). Re-tested: 4 negative scenarios + positive.

PR #381, opened 2026-07-22
fix: #374 web-scraping shell-blind probe + #375 pitfall growth reconciliation

### LINKED ISSUE #374: [BUG] headless-web-scraping: sufficiency probe reports success on a JavaScript app shell
## Summary

`skills/headless-web-scraping/SKILL.md` teaches a "quick probe" for deciding whether the cheap
fetcher tier is sufficient. The check it teaches passes on exactly the page where that tier
failed, so it converts a detectable failure into a confident false pass. A spec that encodes the
wrong check is worse than no spec, because it stops the agent from looking further.

## Findings

1. **[HIGH]** `skills/headless-web-scraping/SKILL.md:61` — the probe is:

   ```python
   if response.status == 200 and response.get_all_text():
       print("Fetcher tier sufficient")
   ```

   A JavaScript app shell satisfies **both** conditions: it returns 200, and its shell markup
   (nav, footer, boilerplate) makes `get_all_text()` non-empty. The probe therefore reports
   "sufficient" precisely when the fetcher returned no real content.

2. **[MEDIUM]** `skills/headless-web-scraping/SKILL.md:71` — the escalation table row says the tell
   is that the "content area is empty". For a shell it is not empty, it is full of chrome.

## Evidence

Reproduced first-hand while extracting a vendor documentation corpus: the host returned
**HTTP 200 with a full Next.js shell for any wrong path**, including a plausible byte count.
Only inspecting the bytes revealed there was no documentation in it. This is the same failure
already recorded in a project memory note (`feedback_spa_shell_not_render`, 2026-05-05) — so it
has now been independently re-derived, which is itself the argument for fixing the spec.

## Suggested fix

The probe cannot judge sufficiency without being told what success looks like. Length thresholds
and app-shell marker lists both fail (shells carry real nav/footer text; framework markers churn).
Make the expectation a required input:

```python
# The probe cannot judge sufficiency without knowing what success looks like.
# Supply a marker you expect on the *rendered* page: a heading, a field label,
# a known row value.
EXPECTED_MARKER = "Quarterly Revenue"   # required, target-specific

fetcher = Fetcher()
response = fetcher.get("https://example.com/target-page")
text = response.get_all_text() if response.status == 200 else ""

if EXPECTED_MARKER in text:
    print("Fetcher tier sufficient")
else:
    print(f"Escalate - status {response.status}, {len(text)} chars, marker absent")
```

Add the reason as a comment: *a JS app shell returns 200 with non-empty text, so
`status == 200 and get_all_text()` passes on exactly the page where the fetcher failed.*

Where no marker is knowable (discovery crawl), fall back to fetching one page with both `Fetcher`
and `DynamicFetcher` and comparing text length — a shell is a small fraction of the rendered page.

## Acceptance Criteria

- [ ] The probe at :61 no longer treats `status == 200` plus non-empty text as sufficiency
- [ ] The replacement requires an expected content marker, with the shell rationale stated inline
- [ ] A no-marker fallback (rendered-vs-raw length comparison) is documented
- [ ] The escalation tell at :71 says "empty **or only chrome/nav copy**"
- [ ] A Common Pitfalls bullet records that a 200 is not evidence of content

## Context

Found during a 3-loop adversarial review of candidate almanac changes, 2026-07-20. Filed as a
defect rather than an enhancement: the current text will actively mislead.

### LINKED ISSUE #375: [BUG] Common Pitfalls sections grow without bound: evolve-skill appends, create-skill caps at 3-6, nothing reconciles them
## Summary

Two authoring skills give contradictory instructions about the `## Common Pitfalls` section, and
nothing reconciles them. The result is unbounded growth in the section that carries the most
valuable content in a skill. This is measurable today: **95 of 371 skills already exceed the
documented cap**, and the worst carry 10.

## Findings

1. **[MEDIUM]** `skills/create-skill/SKILL.md:208` and `:218` specify **"3-6 pitfalls"**.
2. **[MEDIUM]** `skills/evolve-skill/SKILL.md:132` says **"New pitfalls go at the end of the
   Common Pitfalls section"** — an append rule with no prune, merge, or eviction counterpart.
3. **[MEDIUM]** Nothing reconciles the two. `.github/workflows/validate-skills.yml:53` enforces
   only a 500-line ceiling, so the sole backstop is total file length — and it fires long after
   the section stopped being readable.

## Measured state

```
95 of 371 skills exceed the 3-6 cap
worst offenders at 10 pitfalls each:
  version-ml-data, setup-service-mesh, setup-local-kubernetes,
  prune-agent-memory, orchestrate-ml-pipeline
```

## Why it matters more than it looks

`## Common Pitfalls` is a required template section (`skills/_template/SKILL.md`) and is where the
corpus actually stores negative results — the tried-and-rejected knowledge that stops an agent
rediscovering a dead end. `skills/analyze-kernel-bottleneck/SKILL.md:303` shows the section at full
strength, carrying a rejected alternative *with the measurement that killed it*. That value decays
if the section becomes an append-only graveyard: the reader stops reading at bullet seven, so the
newest entries — the ones added because someone just got burned — are the least likely to be seen.

Cost note: every added line is multiplied across the English file plus its translations, against
the 500-line ceiling. Unbounded append is the most expensive growth shape in this repo.

## Suggested fix

One rule in `evolve-skill`, next to the append instruction:

> Above 6 pitfalls, do not append. Drop or move to `references/` the pitfall with the lowest
> rediscovery cost — the one a competent agent would most likely avoid unaided. Pitfalls are
> ranked by what they save, not by age.

Optionally add pitfall-count to `review-skill-format` as a reported metric so the cap is visible
rather than aspirational.

## Acceptance Criteria

- [ ] `evolve-skill` carries an eviction rule, not only an append rule
- [ ] The rule states the ranking criterion (rediscovery cost), not just a number
- [ ] `create-skill`'s 3-6 cap and `evolve-skill`'s growth rule no longer contradict
- [ ] `review-skill-format` reports the pitfall count so drift is observable
- [ ] Decide and record whether the 95 existing over-cap skills are grandfathered or scheduled

## Context

Found during a 3-loop adversarial review, 2026-07-20, while evaluating (and ultimately rejecting)
a proposal to add a new "negative results" section type. The reviewer's point was that the section
already exists and works — its problem is growth management, not absence.

pjt222:
## Summary

Two skill-spec bugs from the 2026-07-20 adversarial review batch.

**#374 — headless-web-scraping.** The sufficiency probe (`status == 200 and get_all_text()`) passes on a JavaScript app shell — 200 plus non-empty nav/footer boilerplate — i.e. on exactly the page where the cheap fetcher tier failed. The probe now requires a target-specific `EXPECTED_MARKER` with the shell rationale inline; a rendered-vs-raw length comparison is documented as the no-marker fallback; the escalation tell reads "empty or only chrome/nav copy"; a Common Pitfalls bullet records that a 200 is not evidence of content.

**#375 — pitfall growth rules.** `evolve-skill` appended without bound while `create-skill` caps at 3-6; 95/371 skills already exceed the cap. `evolve-skill` now carries an eviction rule (above 6, drop the pitfall with the lowest rediscovery cost — ranked by what they save, not by age), `create-skill` states the cap holds for the life of the skill, and `review-skill-format` reports the pitfall count (above-cap = SUGGEST, grandfathered).

## Acceptance criteria mapping

#374: probe no longer treats 200+text as sufficiency ✔ / replacement requires marker with rationale ✔ / no-marker fallback documented ✔ / escalation tell widened ✔ / pitfalls bullet added ✔

#375: eviction rule ✔ / ranking criterion stated (rediscovery cost) ✔ / cap and growth rule reconciled from both sides ✔ / review-skill-format reports count ✔ / grandfathering decision recorded in commit + rule text ✔

## Self-consistency note

Adding the #374 bullet would have pushed headless-web-scraping to 7 pitfalls — breaching the cap #375 fixes. Applied the new eviction rule to it: evicted **No rate limiting**, whose content is fully taught by Step 5's ethical-scraping checklist in the same skill (rediscovery cost ~zero). Count stays at 6.

## Validation

- [x] `check-content-style.js --added origin/main`: 0 blocking errors
- [x] `validate:line-endings`: OK
- [x] All four files well under the 500-line cap
- [x] Pitfall counter self-test: headless-web-scraping 6, version-ml-data 10 (matches issue measurement)

Closes #374, closes #375

### DIFF
--- a/skills/create-skill/SKILL.md
+++ b/skills/create-skill/SKILL.md
@@ -215,6 +215,8 @@ Each item must be objectively verifiable. "Code is clean" is bad. "`devtools::ch
 
 Draw from real experience. The best pitfalls are ones that waste significant time and are non-obvious.
 
+The 3-6 cap holds for the life of the skill: when a later evolution wants a seventh pitfall, [evolve-skill](../evolve-skill/SKILL.md) evicts the pitfall with the lowest rediscovery cost instead of appending past the cap.
+
 **Expected:** 3-6 pitfalls, each with a bold name, a description of what goes wrong, and how to avoid it.
 
 **On failure:** If pitfalls feel generic ("be careful with X"), make them specific: name the symptom, the cause, and the fix. Draw from actual failure scenarios encountered during development or testing.
--- a/skills/create-skill/SKILL.md
+++ b/skills/create-skill/SKILL.md
@@ -13,7 +13,7 @@ license: MIT
 allowed-tools: Read Write Edit Bash Grep Glob
 metadata:
   author: Philipp Thoss
-  version: "1.2"
+  version: "1.3"
   domain: general
   complexity: intermediate
   language: multi
--- a/skills/evolve-skill/SKILL.md
+++ b/skills/evolve-skill/SKILL.md
@@ -129,7 +129,7 @@ Follow these editing rules:
 - Preserve all existing sections — add content, don't remove sections
 - Keep step numbering sequential after insertions
 - Every new or modified step must have both Expected and On failure
-- New pitfalls go at the end of the Common Pitfalls section
+- New pitfalls go at the end of the Common Pitfalls section — but above 6 pitfalls, do not append. Drop (or move to `references/EXAMPLES.md`) the pitfall with the lowest rediscovery cost: the one a competent agent would most likely avoid unaided, e.g. one whose content the skill's own Procedure already teaches. Pitfalls are ranked by what they save, not by age. Skills already over the cap are grandfathered — apply the eviction rule the next time the skill is evolved, not as a bulk retro-edit
 - New related skills go at the end of the Related Skills section
 
 #### For Variants
--- a/skills/evolve-skill/SKILL.md
+++ b/skills/evolve-skill/SKILL.md
@@ -12,7 +12,7 @@ license: MIT
 allowed-tools: Read Write Edit Bash Grep Glob
 metadata:
   author: Philipp Thoss
-  version: "1.2"
+  version: "1.3"
   domain: general
   complexity: intermediate
   language: multi
@@ -129,7 +129,7 @@ Follow these editing rules:
 - Preserve all existing sections — add content, don't remove sections
 - Keep step numbering sequential after insertions
 - Every new or modified step must have both Expected and On failure
-- New pitfalls go at the end of the Common Pitfalls section — but above 6 pitfalls, do not append. Drop (or move to `references/EXAMPLES.md`) the pitfall with the lowest rediscovery cost: the one a competent agent would most likely avoid unaided, e.g. one whose content the skill's own Procedure already teaches. Pitfalls are ranked by what they save, not by age. Skills already over the cap are grandfathered — apply the eviction rule the next time the skill is evolved, not as a bulk retro-edit
+- New pitfalls go at the end of the Common Pitfalls section — but if appending would push the count above 6, do not append: evict the pitfall with the lowest rediscovery cost instead, the one a competent agent would most likely avoid unaided. An eviction must name the Procedure step or section that already teaches the evicted content — if none does, first fold the pitfall's counterintuitive fact into the step that should carry it, then evict. Pitfalls are ranked by what they save, not by age. Skills already over the cap are grandfathered: evict when a later evolution adds a pitfall, not as a bulk retro-edit
 - New related skills go at the end of the Related Skills section
 
 #### For Variants
--- a/skills/headless-web-scraping/SKILL.md
+++ b/skills/headless-web-scraping/SKILL.md
@@ -52,23 +52,35 @@ Determine which scrapling fetcher matches the target site's defenses.
 # 2. StealthyFetcher — Cloudflare/Turnstile, TLS fingerprint checks
 # 3. DynamicFetcher  — JS-rendered SPAs, click/scroll interactions
 
-# Quick probe: try Fetcher first, escalate on failure
+# Quick probe: try Fetcher first, escalate on failure.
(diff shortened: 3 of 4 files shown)


pjt222:
## Review trail

**Copilot**: quota-blocked ("unable to review... reached their quota limit") — standing fallback applied.

**Adversarial verify workflow** (2 agents, one per issue, refute-mandate): 1 NIT on #374 (EXPECTED_MARKER missing from Inputs — folded in f0eefd47); on #375: 2 MEDIUM (format-blind pitfall counter; missing version bumps), 3 LOW (grandfather-clause ambiguity, above-6 off-by-one, Step 4.5 translations flag), 1 NIT (Validation checklist) — all folded in 414a9674.

**advocatus-diaboli** (design-decision attack, Copilot substitute): no blockers; F1 (eviction rule under-determined + EXAMPLES.md clause illusory) folded in 414a9674; F2 (grandfathering leaves static acute tail unaddressed) → filed #382 with evolve-recency data (10 of 16 acute-tail skills static since 2026-02-19); F3 (429-absence signal lost by eviction) + F4a/4b (marker brittleness, fallback threshold/blind spot) folded in aebbda5d; F5 (advisory-only count check) documented in 414a9674.

Notable: the verify workflow's counter test found 25 skills whose pitfalls are numbered/table-format — the dash-only counter under-counted exactly where the drift lives, and the corrected counter grows the measured over-cap set (see #382 comment). Also surfaced: `write-helm-chart` has no Common Pitfalls section at all (filed separately).

Local gates after folds: content-style 0 blocking, all four files under 500 lines, counter verified against dash (6/10), numbered (6/10), table (0→manual), and no-section formats.

PR #384, opened 2026-07-22
feat: #371 i18n keep-in-English parity gate + 39-file allowed-tools catch-up

### LINKED ISSUE #371: i18n: CI parity gate for keep-in-English frontmatter fields (allowed-tools drift class) — PR #368 follow-up
## Summary

The 20-file `allowed-tools` drift fixed by PR #368 persisted for ~2 months because **no CI gate enforces i18n keep-in-English field parity**. `i18n/README.md` declares `name`, `allowed-tools`, tags, domain, language (and code blocks) as keep-in-English, but nothing checks that these fields in `i18n/<locale>/skills/<id>/SKILL.md` equal the English source. When a source's tool grant changes (as #354/PR #358 did), all locale snapshots silently desync.

Identified as the root-cause finding of PR #368's adversarial review. PR #368 is the manual catch-up; this gate is the durable fix (per the durable-structural-fixes-over-one-offs standing rule).

## Proposal

Add a parity check (validate-skills.yml step or a `scripts/` checker) that for every translated skill compares keep-in-English frontmatter fields against the English source:

- `allowed-tools` (the field that actually drifted)
- optionally `metadata.tags` / domain / language, if shape variance permits a robust comparison (note: pilot-era Shape B files carry locale fields at top level — the checker must tolerate both shapes or normalize first)

Warn-only initially (same rollout pattern as A9/#356), promote to blocking once the corpus is clean.

## Acceptance criteria

- [ ] Checker compares `allowed-tools` in all i18n skill copies against English; warn on mismatch
- [ ] Runs in CI on PRs touching `skills/**` or `i18n/**/skills/**`
- [ ] Current corpus passes (post-#368 it should — verify)
- [ ] Decision recorded on whether to extend to other keep-in-English fields

## Provenance

Backlog triage sweep 2026-07-19 → PR #368 adversarial review, Finding 1 (root cause). Filed per the standing file-issues-for-gaps order.

pjt222:
## Summary

Root-cause fix for the drift class behind PR #368: no gate enforced the keep-in-English `allowed-tools` field in translations, so source tool-grant changes silently desynced all locale snapshots.

**Commit 1** — `scripts/check-i18n-frontmatter-parity.js` + warn-only CI step in validate-skills.yml + `npm run validate:i18n-parity`. Scoped line extraction rather than full YAML parse (tolerates pilot-era Shape A/B variance; malformed unrelated fields can't mask the comparison; frontmatter-only scope keeps `allowed-tools:` inside body code examples — the #369 false-positive class — unmatched). Handles inline and block-list field forms; reports MISMATCH / MISSING / ORPHAN.

**Commit 2** — 39-file catch-up. The issue's premise ("current corpus passes post-#368") was **false on verification**: 39 more drifted files existed beyond #368's 20 (stale multi-locale snapshots, mostly `Read Grep Glob WebFetch WebSearch` vs narrowed sources; one `none` vs `Read`). All fixed to the English value verbatim, single-line diffs.

## Acceptance criteria mapping

- Checker compares allowed-tools in all i18n skill copies vs English, warns on mismatch ✔ (3,572 files vs 368 sources)
- Runs in CI on PRs touching `skills/**` or `i18n/**/skills/**` ✔ (validate-skills.yml already has exactly these path filters)
- Current corpus passes ✔ — after the catch-up this PR ships; the premise that it already passed was verified false
- Decision on extending to other keep-in-English fields ✔ — allowed-tools only for now; survey found pre-existing paraphrase drift in `tags` (96 files) and `language` (32) that needs its own triage/catch-up before gating — filing as follow-up

## Verification

- Negative-tested: seeded mismatch → exit 1 + `FAIL [MISMATCH]` line; `--warn` → exit 0; restored → clean (the A9 lesson: a check that has never seen a failure is unproven)
- Blocking-mode exit 0 on the fixed corpus; warn-mode wired in CI per the issue's A9/#356 rollout choice; promotion = drop `--warn` (noted inline in the workflow)
- line-endings gate OK; new script is pure node builtins (no deps)

Closes #371

### DIFF
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -117,3 +117,9 @@ jobs:
 
       - name: Check translation freshness (warn only)
         run: node scripts/check-translation-freshness.js --warn
+
+      # Keep-in-English frontmatter parity (#371, warn-only rollout like A9/#356).
+      # Corpus verified clean at introduction (39-file catch-up in the same PR);
+      # promote to blocking by dropping --warn once the warn stream stays quiet.
+      - name: Check i18n frontmatter parity (warn only)
+        run: node scripts/check-i18n-frontmatter-parity.js --warn
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -118,8 +118,11 @@ jobs:
       - name: Check translation freshness (warn only)
         run: node scripts/check-translation-freshness.js --warn
 
-      # Keep-in-English frontmatter parity (#371, warn-only rollout like A9/#356).
-      # Corpus verified clean at introduction (39-file catch-up in the same PR);
-      # promote to blocking by dropping --warn once the warn stream stays quiet.
-      - name: Check i18n frontmatter parity (warn only)
-        run: node scripts/check-i18n-frontmatter-parity.js --warn
+      # Keep-in-English frontmatter parity (#371). Ships BLOCKING, not warn-only:
+      # the corpus was verified clean at introduction (39-file catch-up in the
+      # same PR), so the issue's promotion precondition was already met, and a
+      # warn stream nobody watches would reproduce the root cause (#371 = two
+      # months of silent drift). Deterministic exact comparison, unlike the
+      # heuristic warn-only A9. Local `npm run validate:i18n-parity` matches.
+      - name: Check i18n frontmatter parity
+        run: node scripts/check-i18n-frontmatter-parity.js
--- a/i18n/de/skills/analyze-prime-numbers/SKILL.md
+++ b/i18n/de/skills/analyze-prime-numbers/SKILL.md
@@ -6,7 +6,7 @@ description: >
   Anwendung des Primzahlsatzes. Verwenden zur Bestimmung der Primalität,
   Faktorzerlegung zusammengesetzter Zahlen und Analyse der Primzahlverteilung.
 license: MIT
-allowed-tools: Read Grep Glob WebFetch WebSearch
+allowed-tools: Read Bash
 metadata:
   author: Philipp Thoss
   version: "1.0"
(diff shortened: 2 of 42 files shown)


pjt222:
## Review trail

**Copilot**: quota-blocked — standing fallback applied (advocatus-diaboli).

**advocatus-diaboli verdict: SHIP with one FIX.** All four attack surfaces:

1. **Direction-of-fix** (English-canonical assumption): ACCEPT, empirically confirmed — for every drifted skill, the 6 newer locales (caveman-*/wenyan-*) already carried the current English value; only the 4 older locales were outliers (scaffold-default `Read Grep Glob WebFetch WebSearch` snapshots gone stale). Even the `heal` anomaly (3 different stale values across old locales) converges on current English. No file held a "right" value against a wrong English source.
2. **Checker edge cases**: ACCEPT — body-fence `allowed-tools:` (create-skill carries one in every locale) excluded by construction; CRLF tolerated; ORPHAN conflation acceptable. One latent blind spot found → **fixed in 383cd9e6**: new EXTRA class flags a translation carrying `allowed-tools` when the English source has none (negative-tested: seeded English-null → EXTRA ×10 locales; restored → exit 0).
3. **CI wiring**: triggers verified sufficient (source-side `skills/**` covered — the exact #371 root-cause path). The FIX: warn-only was a dead gate (unwatched stream = the non-enforcement that caused #371), and the corpus is already clean, so blocking is green today and actually enforces. **Promoted to blocking in 383cd9e6** — a deliberate, documented deviation from the issue's warn-first wording: the issue's own promotion precondition ("once the corpus is clean") is satisfied by this PR, and unlike the heuristic A9 this check is deterministic. CI going green on this PR is itself the executable proof of the corpus-clean claim the reviewer asked for.
4. **Scope honesty** (#369 false-positive class): ACCEPT, verified against the real create-skill test case.

Follow-ups: tags/language extension already filed as #385; the English-null blind spot was fixed here rather than filed.

Issue #386, opened 2026-07-22
viz: glyph for restore-diagram-legibility (new skill, #376)


pjt222:
New skill `restore-diagram-legibility` (PR for #376) needs its pictogram glyph for the force-graph visualization, per the create-glyph procedure. Concept suggestion: a diagram node grid shrinking into a magnifier, or a wide canvas folding into stacked rows. Batch with the next glyph wave; audit-icon-pipeline will flag it until then.

Filed per the file-issues-for-gaps standing order.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

pjt222:
PR #387's verifier noted the auto-generated README claim ("369 R-generated skill icons") overstates by one until this glyph lands — the count derives from the registry, not the icon inventory. Lands as accurate again once the glyph ships.

PR #387, opened 2026-07-22
feat: content batch #376-#379 — diagram legibility skill + security-review merges

### LINKED ISSUE #376: [NEW SKILL] Restore legibility to a generated graph diagram that outgrew its canvas
## Summary

No existing skill covers making an auto-generated graph diagram readable once it has outgrown its
canvas. `generate-workflow-diagram` covers producing one (zero legibility content: no font, scale,
direction or wrapping coverage); `render-publication-graphic` is DPI/colour-profile/print-export
only. The gap is the middle case: the diagram generates fine, and is unreadable.

## The problem shape

A generated map rendered at **6.9% scale** in its display column, putting nominal 12px labels at
**0.83px**. Fullscreen was a 2x improvement against a ~15x deficit. The content was fine — forced
to 1:1 the labels were crisp — so this is a delivery failure, not an authoring one.

## Levers, as measured unit reductions

| lever | cause it addresses | measured |
|---|---|---|
| source-level label wrapping | node width tracks the longest unbroken line | 2685 -> 901 units |
| invisible chain links between disconnected components | engine puts unconnected nodes in one wide row | 3292 -> 881 units |
| per-cluster splitting | one canvas holding a whole system | median 0.8px -> 25.9px effective |
| layout direction, chosen by rendering BOTH and measuring | direction is shape-dependent | 2682 -> 1311 units |

Result across 39 diagrams: median effective label 13.7px -> 25.9px; 30/39 -> 37/39 at 10px or better.

## The rejected levers matter more than the accepted ones

1. **Raising the render font is a no-op.** Layout engines measure in text units, so the canvas
   grows with the font. Measured 5.2px -> 8.3px effective for a much larger file. This is a
   property of the Sugiyama/Dagre layout family, not of any one tool version.
2. **Predicting layout direction from topology fails.** A star-shaped test (one hub holding most
   edges -> flip direction) helped the two diagrams still too wide but **regressed 13 of 38**,
   dropping the median from 25.9px to 16.3px. Direction can only be measured, not predicted —
   render both, keep the narrower.
3. **Sub-splitting by node count is the wrong axis.** The illegible clusters had a median of 6
   nodes, identical to the legible ones. One was 3292 units wide with 6 nodes and zero edges.

## Proposed description line

```yaml
description: >
  Restore legibility to an auto-generated graph diagram that has outgrown its
  canvas - rendering at single-digit percent scale with sub-pixel labels.
  Covers the levers that measurably reduce layout units (source-level label
  wrapping, invisible chain links so unconnected nodes stop spreading into one
  row, per-cluster splitting, layout direction picked by rendering both and
  measuring) and the rejected ones (raising the render font is a no-op because
  layout engines measure in text units; a topology heuristic for direction
  regressed 13 of 38 diagrams). Use when a generated Mermaid/Graphviz/DOT
  diagram is too small to read; not for hand-authoring a small diagram.
```

## Durability condition (from review)

The core is engine-general: layout engines measure in text units, so the lever is the text, not
the font. Disconnected-component packing and cluster splitting are equally general. What rots is
syntax (`~~~`) and the specific numbers. **Ship the numbers as one measured instance, never as
targets**, and name each construct generically with the tool spelling in parentheses.

## Acceptance Criteria

- [ ] Skill scoped to generated graph diagrams that outgrew their canvas, not diagram aesthetics
- [ ] Opens with the measurement (scale factor / effective label size), so the reader can confirm the situation matches before applying anything
- [ ] Leads with the two rejected levers, each with the measurement that killed it
- [ ] States levers as measured unit reductions from one instance, not as targets
- [ ] Names constructs generically, with the Mermaid spelling in parentheses
- [ ] Carries a Common Pitfalls bullet: a rejected approach earns its space only if it ships the measurement that killed it; "we tried X, it did not help" is not a negative result
- [ ] Cross-referenced from `generate-workflow-diagram`

## Context

From a 3-loop adversarial review, 2026-07-20. Two loops independently confirmed this is not a
duplicate. Conditioned on the durability rule above; without it the reviewer's verdict was that it
belongs in a guide rather than a skill.

### LINKED ISSUE #377: [MERGE] monitor-binary-version-baselines: string-pool adjacency artifacts, and why the obvious discriminator is unsound
## Summary

A compiled/bundled binary pools string literals back-to-back with no separator, so a greedy
extract (`grep -oE 'prefix_[a-z0-9_]+'`) runs past the end of one literal into the head of the
next and manufactures an identifier that never existed. Those phantoms then look like real
version-to-version churn. `monitor-binary-version-baselines` has zero coverage of this
(grep: 0 hits for string pool / adjacency / pooled), and it is not a restatement of the existing
"minified names rename per rebuild" pitfall — that is churn in *real* symbols; this manufactures
symbols that were never in the binary.

## Why it is worth a spec

The phantom differs from the real literal **and** the neighbouring literal differs between builds,
so a byte-stable identifier appears as an ADD on one side and a REMOVE on the other. In one
window this shipped false premises to 5 of 16 analysis agents before being caught.

## The important part: the obvious discriminator is UNSOUND as a suppression rule

The natural test — strip 1-2 trailing chars, and if the remainder appears in **both** builds call
it an artifact — must **not** be used to suppress. Real names routinely extend existing ones:

> Build A has `flag_penguin`. Build B genuinely adds `flag_penguins`. Strip one char, remainder
> present in both, and a **real new identifier is silently dropped**.

Singular/plural, `_v2`, `_enabled` -> `_enabled_b` all hit this, and in a flag namespace they are
common rather than exotic. Used as a verdict it suppresses exactly the class of finding the audit
exists to catch. This was caught in review *after* the rule had already been implemented as a
suppression in a live audit tool — it was striking real candidates through and excluding them from
the verified counts.

## Suggested wording

> Pooled literals sit back-to-back, so a greedy extract can run past one literal into the next and
> manufacture an identifier that never existed.
>
> **Triage:** strip 1-2 trailing chars. If the remainder exists in both builds the candidate is
> **UNRESOLVED, not refuted** — a genuine new name is often an extension of an old one.
>
> **Resolve** it one of two ways: (1) *neighbour check* — are the stripped trailing chars the
> opening chars of the adjacent pooled literal? then it is an adjacency artifact; (2)
> *reference-site check* — does the candidate appear anywhere other than the string pool (a real
> read/call site)? then it is real, whatever the trim says.
>
> Never fold unresolved candidates into a "no new identifiers" headline. Report them as a separate
> count so they cannot vanish.

## Placement

- `skills/monitor-binary-version-baselines/SKILL.md` Step 6 caveats (where extraction happens) — ~6 lines
- one Common Pitfalls bullet — ~2 lines
- one-line cross-reference from `sweep-flag-namespace`, where the harm actually lands

## Acceptance Criteria

- [ ] Step 6 describes pooling and the phantom-identifier mechanism
- [ ] The trim test is stated as **triage** producing UNRESOLVED, never as a verdict
- [ ] Both resolution checks (neighbour, reference-site) are documented
- [ ] An explicit warning that unresolved candidates must not be counted as zero
- [ ] Common Pitfalls bullet added
- [ ] Cross-reference from `sweep-flag-namespace`

## Context

From a 3-loop adversarial review, 2026-07-20. Loop 3 refuted the discriminator as originally
proposed; the correction has already been applied to the originating project's tooling. Filing the
corrected form so the unsound version is not what gets generalised.

### LINKED ISSUE #378: [MERGE] Redaction allowlists can be vendor-doc-derived, with a per-token caveat
## Summary

A redaction boundary is usually built as a deny-list of known-sensitive tokens. That design is
structurally always behind: it catches only what someone remembered to add, and it *fails silently*
— a clean pass means "none of my listed patterns matched", not "nothing leaked".

The inversion is to scan for anything shaped like an internal identifier and **subtract what the
vendor has published**. That fails closed, and it makes the boundary mechanical rather than a
judgment call. Measured on one corpus: **134 of 370 internal environment-variable names had since
been documented by the vendor** and were publishable verbatim; 494 documented names formed the
allowlist.

Note this is a **token** allowlist and is distinct from the **file** allowlist already in
`redact-for-public-disclosure` (`tools/public-allowlist.txt`) — different axis, not a duplicate.

## The caveat, which is load-bearing

"The vendor documented it, therefore we may publish it" is unsafe unscoped. Four attacks:

1. **Scope** — the doc licenses the *name*, not your *claim about it*. Publishing `FOO_DISABLE_X`
   is fine; publishing "and setting it to 0 disables the sandbox" is a new disclosure.
2. **Combination** — five documented names plus your ordering is an undocumented architecture map.
3. **Drift** — the vendor removes the page and the justification evaporates.
4. **Provenance** — publishing the name is safe; publishing that you recovered it from a binary at
   a given offset is not.

Wording to carry in both places:

> Allowlisting is **per-token and licenses the name only** — never the value, behaviour,
> co-occurrence, or provenance attached to it. Pin a dated vendor URL per token and re-verify at
> disclosure time, not once at allowlist time.

## Also worth stating: the boundary moves in both directions

A **shrinking** deny-list is a signal, not noise — it means the vendor documented something, which
*relaxes* what may be published. Re-derive the allowlist on a cadence rather than once.

## Placement

- `skills/enforce-redaction-gate/SKILL.md` Step 6 ("Narrow False Positives, Never Suppress") — ~2 lines on doc-derived narrowing plus the re-derivation cadence
- `skills/redact-for-public-disclosure/SKILL.md` Step 1 (the fact inventory) — ~3 lines, as a fact-classification category

Deliberately **not** a new Step: the judgment of what is publishable belongs in the fact inventory,
not inside an executable checker, and `enforce-redaction-gate` is already 204 lines.

## Acceptance Criteria

- [ ] Step 6 notes that the allowlist may be derived from vendor documentation, and must be re-derived on a cadence
- [ ] The per-token caveat appears in both files, in full
- [ ] The token-allowlist vs file-allowlist distinction is stated so the two are not conflated
- [ ] A shrinking deny-list is described as a signal worth investigating
- [ ] No new Step added to `enforce-redaction-gate`

## Context

From a 3-loop adversarial review, 2026-07-20. Loop 1 confirmed this is not a duplicate of the
existing file allowlist; loop 3 supplied the caveat after attacking the rule as unsafe unscoped.

### LINKED ISSUE #379: [MERGE] enforce-redaction-gate Step 3: re-parse the redacted artifact with its own parser
## Summary

Redaction runs *before* the parser. A token replaced in **identifier position** produces
syntactically broken output that no visual diff and no token-level gate will surface — the gate
passes, the artifact is corrupt, and the corruption is discovered downstream or in production.

This belongs in `enforce-redaction-gate` **Step 3** (the structure-aware / AST-checker tier,
`:101`) — same phase, same tool class. Not a new Step: it is a three-line check and the skill is
already 204 lines.

## Findings

1. **[MEDIUM]** Replacing a token inside a JavaScript object key produced `trigger-tool:` and
   `[sandbox-classifier-gate]_fail_open_inversion:` — both syntactically invalid. `node --check`
   caught it instantly; a diff review had not.
2. **[MEDIUM]** The same replacement silently broke three `data-count="<key>.<field>"` lookups,
   because the key in the markup and the key in the data file were rewritten inconsistently. The
   page rendered with empty figures and no error.
3. **[MEDIUM]** In a diagram source, a hyphenated replacement in node-ID position produced
   `opus47-launch`, which the renderer stopped treating as an identifier — **silently dropping 2
   nodes and 2 edges** from the published graph. Detected only by comparing node counts between
   the private and public builds (223/96 vs 221/94).

## Suggested wording

> Redaction runs before the parser, so a token replaced in identifier position produces
> syntactically broken output that no visual diff surfaces. Re-parse the **redacted** artifact with
> its own parser — `node --check`, `mmdc`, `jq`, a YAML load — not the original. Where the artifact
> references its own identifiers (data-attribute lookups, graph edges), assert that every reference
> still resolves after redaction, and that structural counts (nodes, edges, keys) match the source.

Generalised rule worth stating once: **a replacement that is safe in prose is dangerous in
identifier position.** Validate the output, not just the token list.

## Placement

`skills/enforce-redaction-gate/SKILL.md` Step 3 (:101) — ~3 lines, plus one Validation checkbox.

## Acceptance Criteria

- [ ] Step 3 instructs re-parsing the redacted artifact with its own parser
- [ ] The prose-vs-identifier-position distinction is stated explicitly
- [ ] Reference-resolution and structural-count checks are covered for self-referencing artifacts
- [ ] One Validation checkbox added
- [ ] No new Step introduced

## Context

From a 3-loop adversarial review, 2026-07-20. Loop 3 cut this down from a proposed new Step, and
corrected the placement in my brief — the AST-checker move is Step 3, not Step 5. The node-ID half
was dropped as already covered by `redact-visualization-for-disclosure`.

pjt222:
## Summary

Four issues from the 2026-07-20 3-loop adversarial review batch, one commit each (+1 scaffold commit).

**#376 — new skill `restore-diagram-legibility`** (visualization, 369th skill). The gap between generate-workflow-diagram (producing) and render-publication-graphic (print delivery): a generated diagram rendering at single-digit percent scale with sub-pixel labels. Honors the issue's durability condition — opens with the deficit measurement, leads with the two rejected levers each carrying the measurement that killed them, states working levers as measured unit reductions (never targets), generic construct names with Mermaid spellings in parentheses, and the negative-results-need-their-measurement pitfall. Registry + symlinks (via sync-discovery-symlinks.sh) + README regen + de/zh-CN/ja/es scaffolds (post-source-commit, so source_commit captures shipped HEAD). Glyph deferred to the next wave: #386.

**#377 — string-pool adjacency artifacts** in monitor-binary-version-baselines Step 6: greedy prefix-extracts manufacture phantom identifiers from pooled literals; the trim test is documented as triage producing UNRESOLVED (never a suppression — real names extend old ones); neighbour + reference-site resolution checks; unresolved candidates get a separate count. Pitfall added with a rule-compliant eviction (`strings -n 4` pitfall — content verbatim in Step 6's body; count stays 8, grandfathered one-in-one-out). Cross-ref from sweep-flag-namespace.

**#378 — vendor-doc-derived token allowlists**: enforce-redaction-gate Step 6 (derivation, cadence, shrinking-deny-list-as-signal) + redact-for-public-disclosure Step 1 (fact-inventory category). The load-bearing per-token caveat appears in full in both files: the doc licenses the name only — never value, behaviour, co-occurrence, or provenance; dated vendor URL per token, re-verified at disclosure time. Token-vs-file allowlist distinction stated in both.

**#379 — re-parse the redacted artifact** in enforce-redaction-gate Step 3: a replacement safe in prose is dangerous in identifier position; re-parse the redacted output with its own parser; reference-resolution + structural-count assertions for self-referencing artifacts. One Validation checkbox; no new Step.

## Validation

- [x] content-style `--added origin/main`: 0 blocking; line-endings OK
- [x] New skill 201 lines; all edited skills far under 500
- [x] Parity gate (new, blocking): 3,576 translated skills green — validated its own scaffolds
- [x] Version bumps on all six touched skills (five 1.0→1.1, generate-workflow-diagram 1.0→1.1)
- [x] total_skills 368→369; README auto-sections regenerated

Translations flagged for re-sync: de, zh-CN, ja, es, caveman-{lite,full,ultra}, wenyan-{lite,full,ultra} for the five edited skills (corpus-wide refresh tracked by #278).

Closes #376, closes #377, closes #378, closes #379


pjt222:
## Review trail

**Copilot**: quota-blocked — standing fallback applied.

**Verify workflow** (4 agents, one per issue): 1 MEDIUM (Step 3 topic-sentence referent contradicted the gate's own data-flow definitions), 5 LOW (corpus arithmetic slip "13 of 38 others"; viz README icon count overstated pending glyph — noted on #386; Step 4.5 translations flag — see below; caveat missing its closing clause; Step 1 category machinery not reconciled with the new paragraph), 2 NIT (semver-table ambiguity — filed separately; shrinking-deny-list off-axis with the shape doctrine). All folded in d53db582 except the two tracked externally.

**advocatus-diaboli** (design attack): no blockers. F1 skill-not-guide ACCEPT (6 of 7 steps carry failable substance; Step 2's unverifiable Expected made failable in d53db582); domain placement `visualization` ACCEPT (workflow-visualization is putior-specific by definition). F2 measured-instance framing ACCEPT with the floor-formula anchor added. F3 — the finding I'd single out: the #377 eviction's "taught verbatim" claim was **literally false**; the inoculating negation ("-n 4 is not more thorough") had been lost, repeating the exact class PR #381's first eviction needed a fold-back for. Kernel restored into Step 6's caveat; the eviction itself stands (the evolve-skill rule mandates one-in-one-out for over-cap skills — restoring the bullet would break the rule shipped in #381). F4 checkbox asymmetry fixed (per-token caveat now has its own Validation checkbox) and the allowlist-vs-suppression tension reconciled in text.

**On Step 4.5 (translations)**: source_commit bumps without content re-translation would fake freshness (the status-yml-lies class), so per PR #381 precedent the re-sync flag rides in the PR body/merge and staleness stays visible to the warn-only freshness check, tracked corpus-wide by #278.

Same eviction-rule meta-note as #381: two independent eviction attempts have now both initially lost the counterintuitive kernel — evidence the fold-first-then-evict wording the rule gained in #381 is the load-bearing half.

Issue #382, opened 2026-07-22
skills: one-time pitfall-eviction pass on the 9-10-pitfall tail (16 skills, mostly static since Feb)


pjt222:
## Summary

PR #381 (#375) added an eviction rule to `evolve-skill` (above 6 pitfalls, evict the lowest-rediscovery-cost one) and grandfathered the 95 existing over-cap skills — eviction applies on next evolve touch. The adversarial review accepted the grandfather decision but flagged its unmeasured assumption: next-touch healing only works if over-cap skills actually get touched.

Measured (2026-07-22): the acute tail does **not** churn. Of the 16 skills at 9-10 pitfalls, 10 were last touched 2026-02-19 (5 months static, the devops/mlops batch), 1 in March, 3 in June:

```
10 pitfalls: version-ml-data, setup-service-mesh, setup-local-kubernetes,
             prune-agent-memory, orchestrate-ml-pipeline, optimize-cloud-costs,
             navigate-dach-bureaucracy, enforce-policy-as-code,
             deploy-ml-model-serving, configure-api-gateway, build-feature-store,
             assess-github-repo-security
 9 pitfalls: register-ml-model, expand-awareness, create-skill, circuit-breaker-pattern
```

Under grandfathering-only these stay maximally unreadable indefinitely, and `review-skill-format` only SUGGESTs on count — no forcing function.

## Proposal

One-time targeted eviction pass on just the >=9 tail (16 skills, not the full 95): per skill, evict down to 6 by the PR #381 rule — lowest rediscovery cost, each eviction naming the Procedure step that already carries the content. Small enough to review each diff individually, which is exactly why the bulk-edit objection to un-grandfathering the full 95 does not apply here.

The 32 skills at exactly 8 and the rest of the 95 at 7 stay grandfathered (readability harm is marginal there; next-touch healing suffices).

## Acceptance criteria

- [ ] Each of the 16 skills reduced to <=6 pitfalls
- [ ] Every eviction names the Procedure step/section that already teaches the evicted content (or gives an explicit lowest-value rationale where none does)
- [ ] No counterintuitive signal lost: where an evicted bullet carried a fact its covering step does not sharply state, the fact is folded into that step first (pattern: PR #381's 429-absence fold)
- [ ] One PR, per-skill commits, so each eviction diff is individually reviewable

## Provenance

advocatus-diaboli review of PR #381, Finding 2 — filed per the file-issues-for-gaps standing order.

[...]

Issue #13, opened 2026-02-16
feat(agents): add skill-reviewer agent with review-domain skills (3 skills)


pjt222:
## Summary

Add a new **skill-reviewer** agent specializing in agentskills.io format review, content updates, and structural refactoring. Expand the existing `review` domain from 6 to 9 skills.

## Agent: `skill-reviewer`

- **Description**: Reviews, updates, and refactors SKILL.md files for agentskills.io compliance and quality
- **Tools**: Read, Write, Edit, Bash, Grep, Glob (needs write access for updates)
- **Model**: sonnet (default)
- **Related agents**: code-reviewer, janitor

## Expanded Domain: `review` (6 → 9 skills)

| Skill | Description |
|-------|-------------|
| `review-skill-format` | Review a SKILL.md file for agentskills.io format compliance, frontmatter validity, and section completeness |
| `update-skill-content` | Update skill content — refresh procedures, add new steps, update references |
| `refactor-skill-structure` | Refactor a skill's structure — extract references, split oversized files, reorganize sections |

## Tasks

- [ ] Create `agents/skill-reviewer.md` from `agents/_template.md`
- [ ] Create `skills/review/review-skill-format/SKILL.md`
- [ ] Create `skills/review/update-skill-content/SKILL.md`
- [ ] Create `skills/review/refactor-skill-structure/SKILL.md`
- [ ] Add agent to `agents/_registry.yml`
- [ ] Add skills to `skills/_registry.yml` under `review` domain, update `total_skills` count
- [ ] Add `.claude/skills/` symlinks for slash command discovery

## Acceptance Criteria

- Agent file follows `agents/best-practices.md` conventions
- All 3 SKILL.md files have valid frontmatter and all required sections
- Each SKILL.md is ≤500 lines
- Registry counts are accurate (review domain: 9 skills)
- Skills cross-reference existing review skills and code-reviewer/janitor agents

PR #402, opened 2026-07-23
sweep: write-helm-chart sections + required-section CI gate, evolve-skill semver table, glyphs/generator hygiene (#383 #388 #390)

### LINKED ISSUE #383: [BUG] write-helm-chart: missing required Common Pitfalls section (and skills-ref validate does not catch it)
## Summary

`skills/write-helm-chart/SKILL.md` has no `## Common Pitfalls` section — one of the six required sections per the template and `review-skill-format`. Surfaced by PR #381's verify workflow while testing the new pitfall counter (the only skill of 369 with the section absent entirely).

Second finding implied: CI (`skills-ref validate` in validate-skills.yml) passes on main, so the required-sections check either is not part of skills-ref validation or tolerates the absence. Worth confirming which when fixing — if skills-ref does not check sections, `review-skill-format` remains the only guard and it runs on-demand, not in CI.

## Acceptance criteria

- [ ] `write-helm-chart` gains a Common Pitfalls section (3-6 items per create-skill)
- [ ] Confirmed whether skills-ref validate checks required sections; finding recorded here

## Provenance

PR #381 verify workflow, counter format-coverage test, 2026-07-22. Filed per the file-issues-for-gaps standing order.

### LINKED ISSUE #388: evolve-skill: Step 5 semver table says 'New step, new pitfall, new table -> Minor: 1.0 -> 2.0' — mislabeled and ambiguous
`skills/evolve-skill/SKILL.md:218` maps "New step, new pitfall, new table" to "Minor: 1.0 → 2.0" — but 1.0 → 2.0 is a **major** bump in semver terms, and repo practice treats content additions as 1.0 → 1.1 (13 skills sit at 1.1; PR #381 and #387 both bumped x.y → x.y+1 for new pitfalls). Either the label is wrong (should be "Major"), or the example is wrong (should be 1.0 → 1.1), and the criterion for which additions are major vs minor is undefined.

Surfaced by PR #387's verify workflow (NIT), which noted the ambiguity reads "as much a defect in evolve-skill's table as in the PR".

## Acceptance criteria

- [ ] Table row label and example agree with each other and with semver terminology
- [ ] The major-vs-minor criterion for content additions is stated (suggestion: breaking procedure changes = major; additive steps/pitfalls/tables = minor)
- [ ] Existing corpus versions are NOT retro-bumped (grandfathered)

Filed per the file-issues-for-gaps standing order.

### LINKED ISSUE #390: viz: generator/comment hygiene — stale glyphs.R header, design-skills grouping, timestamp-only regeneration churn
Three pre-existing nits surfaced by PR #389's adversarial review (none merge-blocking):

1. `viz/R/glyphs.R` header says "Maps each of 350 skillIds" — stale vs 369; make it count-free or derive it.
2. `create-glyph` / `enhance-glyph` are listed under the `# ── visualization` comment block in glyphs.R while registry/manifest place them in `design` — regroup or note the comment blocks are organizational, not authoritative.
3. Generators are non-idempotent: `viz/js/colors-generated.js`, `viz/public/data/palette-colors.json`, `viz/public/data/team-icon-manifest.json` rewrite a `Generated:` timestamp even when content is semantically unchanged, producing pure-churn diffs in every icon PR. Skip the write when content (minus timestamp) is stable.

Provenance: advocatus review of PR #389, 2026-07-22. Filed per the file-issues-for-gaps standing order.

pjt222:
## Summary

Three small issues from the 2026-07-22 review trail, one commit each, plus the durable gate that would have caught the first one.

### #383 — write-helm-chart missing three required sections

`skills/write-helm-chart/SKILL.md` ended at Step 6 with no Validation, Common Pitfalls, or Related Skills. Added all three; version `1.0 → 1.1`.

The pitfalls are chosen for non-obviousness rather than coverage — each names a failure that survives the usual check: whitespace chomping renders YAML that passes `helm lint` and fails at apply; an empty `image.tag` resolves to a mutable `latest` so rollback stops meaning anything; hook Jobs are not release-managed, so a completed Job is neither rolled back nor removed and blocks re-install on a name collision; a stale `version` leaves repositories serving the previous chart even after `appVersion` moves; `charts/` is not refreshed by `helm package`; and values files deep-merge rather than replace.

### #383 second AC — why CI stayed green, and the fix

Verified empirically rather than assumed:

```
$ skills-ref validate skills/write-helm-chart
Valid skill: /mnt/d/dev/p/agent-almanac/skills/write-helm-chart
```

`skills-ref` 0.1.0 (the SHA this workflow pins) validates the agentskills.io frontmatter contract, **not** this repo's six-section convention. `review-skill-format` was the only guard and it runs on demand, never in CI.

New `scripts/audit-skill-sections.js` closes that, exposed as `npm run validate:skill-sections` and wired into `validate-skills.yml`. It is promoted from a throwaway analysis script per the extract-reusable-tooling rule, and two parsing details in it are load-bearing — both found by verifying its own output instead of trusting it:

- **Prefix heading match.** The corpus writes `## Validation` (334), `## Validation Checklist` (38), and `## Validation Checks` (1). An equality check reported **36 false positives** before this was fixed.
- **Fenced-code exclusion.** `create-skill` and `evolve-skill` embed a ` ```markdown ` fence containing a literal `## Common Pitfalls` template. A fence-blind parser locks onto the example, then walks past it collecting bullets until the next real heading — it read `create-skill` as 1 pitfall when its real section has 6.

It also counts both supported list formats (dash and numbered); a dash-only counter reads 0 for the ~24 numbered-list skills, which is exactly how the original #382 tail measurement under-counted by 8.

With write-helm-chart fixed, the gate is green corpus-wide (369 skills, 0 reported). Commits are ordered fix-then-gate so bisect stays clean.

### #388 — evolve-skill semver table

The table mapped "New step, new pitfall, new table" to `Minor: 1.0 → 2.0` — a major bump with a minor label. The row above was wrong the other way, calling `1.0 → 1.1` a "Patch" bump although skill versions are two-part `MAJOR.MINOR` with no patch component. Both rows now agree with their examples and with corpus practice.

The undefined major-vs-minor criterion is now stated and anchored to **caller breakage, not edit size**: minor when someone following the old version still gets a correct result, major when the change invalidates how the skill was used. Existing versions stay grandfathered. Also fixed the Step 3 decision matrix, which offered "Patch or minor" and would have contradicted the new text.

No pitfall was added, so the over-cap eviction rule at `evolve-skill:132` does not fire — it stays at 7, grandfathered. Version `1.3 → 1.4`.

### #390 — glyphs.R and generator hygiene

Header is now count-free (claimed 350, registry is at 369). `create-glyph` and `enhance-glyph` moved from the `visualization` comment block to `design`, matching the registry, with both block counts corrected (design 4→6, visualization 7→5) and a header note recording that these blocks are a reading aid while the registry is authoritative.

`export_palette_js` and `export_palette_json` now skip the write when only the timestamp would change, via a shared `content_unchanged()` helper — no more pure-churn diffs in every icon PR.

**Premise correction:** the issue named `viz/public/data/team-icon-manifest.json` as a third churning generator. `build-icon-manifest.js` emits no timestamp field and the manifest carries none — only the two palette outputs churned.

## Verification

- `npm run validate:skill-sections` → 0 reported across the full 369-skill corpus
- `Rscript -e "parse(...)"` clean on both `glyphs.R` and `palettes.R`
- `package.json` parses; `write-helm-chart` at 441 lines, under the 500 cap; LF clean

Closes #383. Closes #388. Closes #390.

### DIFF
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -2,9 +2,9 @@ name: Validate Skills
 on:
   push:
     branches: [main]
-    paths: ['skills/**', 'i18n/**/skills/**']
+    paths: ['skills/**', 'i18n/**/skills/**', 'scripts/audit-skill-sections.js', '.github/workflows/validate-skills.yml']
   pull_request:
-    paths: ['skills/**', 'i18n/**/skills/**']
+    paths: ['skills/**', 'i18n/**/skills/**', 'scripts/audit-skill-sections.js', '.github/workflows/validate-skills.yml']
   workflow_dispatch:
 
 permissions:
@@ -118,10 +118,6 @@ jobs:
             done
           done
           exit $failed
-
-      - uses: actions/setup-node@v7
-        with:
-          node-version: '24'
 
       - name: Install dependencies
         run: npm ci
--- a/.github/workflows/validate-skills.yml
+++ b/.github/workflows/validate-skills.yml
@@ -26,6 +26,10 @@ jobs:
         with:
           python-version: '3.12'
 
+      - uses: actions/setup-node@v7
+        with:
+          node-version: '24'
+
       # Pinned to commit SHA for reproducibility — update periodically
       - name: Install skills-ref
         run: |
@@ -44,6 +48,13 @@ jobs:
             fi
           done
           exit $failed
+
+      # skills-ref validates the agentskills.io frontmatter contract only — it
+      # passes a SKILL.md with no Validation, Common Pitfalls, or Related Skills
+      # section at all (verified against write-helm-chart, issue #383). The
+      # repo's six-section convention needs its own gate.
+      - name: Check required sections
+        run: node scripts/audit-skill-sections.js --missing
 
       - name: Check line counts
         run: |
--- a/guides/creating-skills.md
+++ b/guides/creating-skills.md
@@ -237,7 +237,7 @@ The key decision during evolution is whether to refine in-place or create an adv
 |---|---|---|
 | Skill identity | Unchanged | New ID: `<skill>-advanced` |
 | File location | Same SKILL.md | New directory |
-| Version | Bump patch/minor | Starts at 1.0 |
+| Version | Bump minor (major only if breaking) | Starts at 1.0 |
 | Registry | No new entry | New entry added |
 | Original skill | Modified directly | Left intact |
 
@@ -245,13 +245,16 @@ The key decision during evolution is whether to refine in-place or create an adv
 
 ### Version Bumping
 
-Update the `version` field in frontmatter to track changes:
+Skill versions are two-part, `MAJOR.MINOR` — there is no patch component:
 
 | Change Type | Version Bump | Example |
 |---|---|---|
-| Typo, wording fix | Patch: 1.0 to 1.1 | Fixed unclear sentence in Step 3 |
-| New step, new pitfall | Minor: 1.0 to 2.0 | Added Step 7 for edge case handling |
-| Restructured procedure | Major: 1.0 to 2.0 | Reorganized from 5 to 8 steps |
+| Typo, wording fix | Minor: 1.0 to 1.1 | Fixed unclear sentence in Step 3 |
+| Additive: new step, new pitfall | Minor: 1.0 to 1.1 | Added Step 7 for edge case handling |
+| Breaking: restructured procedure | Major: 1.0 to 2.0 | Reorganized from 5 to 8 steps |
+
+The criterion is whether existing callers break, not how much text changed. See
+[evolve-skill](../skills/evolve-skill/SKILL.md) Step 5 for the full rule.
 
 ## Reviewing Skills
 
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "validate:i18n-parity": "node scripts/check-i18n-frontmatter-parity.js",
     "validate:content-style": "node scripts/check-content-style.js --all",
     "validate:line-endings": "node scripts/check-line-endings.js",
+    "validate:skill-sections": "node scripts/audit-skill-sections.js --missing",
     "translation:status": "node scripts/generate-translation-status.js",
     "translate:scaffold": "bash scripts/translate-content.sh",
     "test": "npm run validate:integrity && npm run check-readmes",
--- a/scripts/audit-skill-sections.js
+++ b/scripts/audit-skill-sections.js
@@ -20,7 +20,7 @@
 
 import { readFileSync, existsSync, readdirSync } from 'fs';
 import { resolve, dirname } from 'path';
-import { fileURLToPath } from 'url';
+import { fileURLToPath, pathToFileURL } from 'url';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const ROOT = resolve(__dirname, '..');
@@ -39,8 +39,8 @@ export const REQUIRED_SECTIONS = [
  * Extract the body of a `## <heading>` section, up to the next `## ` heading.
  *
  * Matches on heading PREFIX, not equality: the corpus uses "## Validation"
- * (334 skills), "## Validation Checklist" (38), and "## Validation Checks" (1)
- * interchangeably, and an equality check reports 39 false positives.
+ * (334 occurrences), "## Validation Checklist" (38), and "## Validation Checks"
+ * (1) interchangeably, and an equality check reports 39 false positives.
  */
 function sectionBody(lines, heading) {
   const wanted = `## ${heading}`.toLowerCase();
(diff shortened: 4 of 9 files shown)


pjt222:
## Review trail: advocatus-diaboli substitute (Copilot quota-blocked)

Copilot returned `COMMENTED` with the quota-limit message again, so the standing fallback ran an adversarial review with the full attack surface: parser correctness, gate fail-open modes, the factual accuracy of every new Helm pitfall, cross-document consistency, and the R idempotency change.

**Verdict: no merge-blocking findings, but it found three factually wrong Helm pitfalls and a latent parser bug.** All actionable findings are now folded in — five follow-up commits.

### Folded in

| # | Severity | Finding | Commit |
|---|---|---|---|
| 3 | MEDIUM | Hook pitfall contradicted upstream Helm docs — the default `before-hook-creation` policy *does* delete the prior hook, so there is no name collision | `c6edf16c` |
| 4 | MEDIUM | "Values deep-merge" omitted the list exception — and this file's own `values-dev.yaml` overrides `ingress.hosts`, which is replaced, not merged | `c6edf16c` |
| 5 | MEDIUM | `image.tag` pitfall described a mechanism that cannot occur given the skill's own Step 3 template — **and missed a real bug in Step 4**, whose hook template lacked `\| default .Chart.AppVersion` while values ships `tag: ""`, rendering `my-app:` | `c6edf16c` |
| 2 | MEDIUM | `fenceMask()` compared fence characters but not run length, so an inner ` ```{r} ` could close an outer ` ````markdown ` fence and phase-flip the mask | `bda5496e` |
| 8 | LOW | Heading-only stub sections satisfied `--missing` | `bda5496e` |
| 12 | LOW | Arg parser failed open on `--min` without a value, on a non-numeric `--min`, on unknown flags, and on nonexistent skill ids | `bda5496e` |
| 11 | LOW | Main-module guard used string concatenation; breaks on paths needing percent-encoding, skipping the CLI and exiting 0 | `bda5496e` |
| 7 | MEDIUM | `paths:` filter excluded the gate's own script, so a PR touching only it would never run it | `bda5496e` |
| 9 | LOW | Duplicate `setup-node` in the same job | `bda5496e` |
| 1 | HIGH | `guides/creating-skills.md` carried the same defective semver table, and fixing only evolve-skill made them disagree on both label and destination version | `034cede2` |
| 10 | LOW | Block-count sweep was incomplete — `general` said 21 with 24 entries, `swarm` said 8 with 9 | `57c8da4b` |
| 15 | NIT | JSON idempotency compared `toJSON()` against a `write_json()` file, coupling the check to a jsonlite internal | `57c8da4b` |
| 14 | NIT | Docstring said "skills" where it meant "occurrences" | `bda5496e` |

### One finding the hardening itself produced

Adding the empty-section check immediately failed CI on `use-graphql-api`, which writes its six pitfalls as a **two-column table** — a third valid format neither the original counter nor the reviewer had considered. Caught by running the hardened gate against the corpus before pushing rather than after. All three formats (dash, numbered, table) are now counted, and `use-graphql-api` reads 6.

### Findings not folded, with reasons

**#6 — six locale copies of `evolve-skill` carry the superseded table.** Correct, and this is a deliberate deviation rather than an oversight. `evolve-skill` Step 4.5 prescribes bumping `source_commit` in each locale file, but doing that without retranslating marks stale content as *fresh* and hides it from `check-translation-freshness.js` — the opposite of what the step is for. Flagging in the commit message (Step 4.5's step 3) is the honest half. Filed separately, along with the tension in Step 4.5 itself.

*Translations flagged for re-sync (evolve-skill, 10 locales): de, zh-CN, ja, es, caveman, caveman-lite, caveman-ultra, wenyan, wenyan-lite, wenyan-ultra. Changed section: Step 5 version table + Step 3 matrix row.*

**#13 — the "13 skills at x.1" claim.** The reviewer correctly declined to verify this from a working tree sitting on the sibling eviction branch. Measured against `main`: **13 skills at exactly `1.1`**, plus one at `2.1` — so "13 at 1.1" is right and my "13 at x.1" phrasing was off by one. Corrected here rather than in history.

While checking it I also confirmed the two-part rule holds universally: the three apparent three-part skill versions (`build-custom-mcp-server`, `scaffold-mcp-server`) are `"1.0.0"` strings inside JSON/JS code examples, not frontmatter. Both files carry `version: "1.0"`.

### Confirmed clean by the reviewer (recorded so it is not re-spent)

`content_unchanged`'s forward reference is safe under every load path (`source()` evaluates the whole file before any exporter is called; six other scripts source `palettes.R` but only reach the colour getters). The JSON idempotency check was **not** a silent no-op even before commit `57c8da4b`. Skipping the write breaks no downstream consumer — `build.sh` is linear with no make-style staleness, and the icon caches key on a content hash computed in memory, not on file mtime. `create-glyph`/`enhance-glyph` really are `design` per the registry. `~~~` fences and `*` bullets: zero in the corpus. ESM/CJS correct, exit-code propagation correct, `--json --missing` preserves exit 1, `--min 9` does not treat `9` as a skill id. `evolve-agent`/`evolve-team`'s three-part "Patch: 1.0.0 → 1.0.1" is **not** a contradiction — agents and teams genuinely use three-part semver; only skills are two-part.

Issue #400, opened 2026-07-23
skills: recheck the #382 eviction pass under Fable — 16 Opus-authored decisions, 24 Opus-only verifications


pjt222:
## Why

The #382 pitfall-eviction pass ran as a Fable 5 workflow that hit the usage-credit ceiling mid-run. Work resumed on Opus 4.8 with fast mode. The pass therefore has **mixed model provenance**, and — importantly — **not a single skill in it was verified by Fable**: all 24 adversarial verify agents in the first run died on credits before returning.

Eviction is a judgment task (which pitfall has the lowest rediscovery cost; whether a counterintuitive kernel survives the fold). The repo has a documented history of that judgment going wrong twice (PRs #381 and #387 both lost an inoculating negation and needed review-driven fold-backs — the reason the fold-first rule exists at `evolve-skill:132`). Model provenance is therefore worth tracking explicitly rather than assuming parity.

This issue marks what needs a Fable recheck when Fable is available again.

## Tier 1 — Opus-authored eviction decisions (recheck first, 16 skills)

These had their *eviction choices* made under Opus 4.8. Highest recheck value.

**Full Opus pass (untouched when credits ran out):**

- [ ] `register-ml-model`
- [ ] `expand-awareness`
- [ ] `create-skill`
- [ ] `circuit-breaker-pattern`
- [ ] `create-2d-composition`
- [ ] `create-3d-scene`
- [ ] `prepare-print-model`
- [ ] `render-blender-output`
- [ ] `render-publication-graphic`
- [ ] `script-blender-automation`
- [ ] `select-print-material`
- [ ] `troubleshoot-print-issues`

**Hybrid — Fable folded the kernels, Opus performed the eviction** (Fable's fold edits were preserved verbatim; only the "which 4 to remove" decision is Opus's):

- [ ] `build-feature-store`
- [ ] `configure-api-gateway`
- [ ] `enforce-policy-as-code`
- [ ] `navigate-dach-bureaucracy`

## Tier 2 — Fable-authored eviction, Opus-verified only (8 skills)

Fable made the eviction decisions and completed them (6 pitfalls, version bumped) before credits ran out; only the adversarial verification is Opus's. Lower risk, but the verify tier is exactly where the two historical kernel losses were caught, so it still deserves a Fable pass.

- [ ] `assess-github-repo-security`
- [ ] `deploy-ml-model-serving`
- [ ] `optimize-cloud-costs`
- [ ] `orchestrate-ml-pipeline`
[...]

Issue #404, opened 2026-07-23
skills: prepare-print-model Step 4 tensile-orientation example was inverted (corrected during #382 eviction repair — needs domain confirmation)


pjt222:
## Context

Surfaced during the #382 pitfall-eviction repair on `skills/prepare-print-model/SKILL.md`.
Evicted pitfall 3 read:

> **Wrong orientation for strength**: Printing tensile parts with layers parallel to load direction creates weak delamination plane

It was the only text in the file naming the FDM anisotropy failure mode. Per the
eviction rule the mechanism was folded into Step 4 (`### 4. Determine Print Orientation`)
before eviction. While folding, the adjacent example turned out to contradict the line
directly above it.

## The defect (pre-existing, not introduced by the eviction)

Old Step 4 text:

```text
- Orient so layer lines run perpendicular to primary load direction
- Example: Bracket under tension -> print vertically so layers stack along load axis
```

"Layers stack along the load axis" puts every layer interface normal to the tensile
load — which is precisely the delamination-prone orientation the first line tells you
to avoid. The example described the weak orientation, not the strong one. The
`**Load direction analysis**` block below used the same ambiguous `layers` noun
(plane? stacking direction?) for the same load cases.

## What was changed in the repair PR

Step 4 now states the mechanism and names the stacking direction explicitly:

```text
- The layer interface is the weak plane, not the material: a part pulled across its
  stacked layers does **not** fail by breaking the extruded bead, it delaminates at
  the bond between layers. ...
- Orient so the layer stacking direction (build Z) runs perpendicular to the primary
  load, keeping the load in-plane where continuous extruded beads carry it
- Example: Bracket under tension -> lay it flat so the tension axis stays in the X/Y
  plane; printing it upright stacks layer interfaces across the load axis and invites
  delamination
```

and the `**Load direction analysis**` block was re-worded from the ambiguous `layers`
to `stack layers` / `stacking direction`.

This is a **physics correction**, not a formatting change, so it is filed here rather
than being left silent inside a pitfall-eviction PR.

## Acceptance criteria

- [ ] A domain reviewer (3d-printing) confirms the corrected orientation guidance:
      tensile load should lie in the X/Y layer plane; build-Z stacking should be
      perpendicular to the primary load.
- [ ] Confirm no other almanac skill repeats the old inverted phrasing
      (`troubleshoot-print-issues`, `select-print-material` checked first).
[...]

PR #406, opened 2026-07-23
skills(#382): one-time pitfall-eviction pass on the >=9 tail — 24 skills to the 6 cap

### LINKED ISSUE #382: skills: one-time pitfall-eviction pass on the 9-10-pitfall tail (16 skills, mostly static since Feb)

pjt222:
## Summary

The one-time eviction pass on the `>=9`-pitfall tail. All 24 skills go to exactly **6** pitfalls, one commit per skill so each diff is individually reviewable (issue AC 4).

Six skills carry a second, follow-up commit. Their first commit captured the file mid-run, before the verify tier's repair round had finished writing — a background `git status` poll loop of mine was contending for `.git/index.lock` and silently starving the commit loop (it kept exiting 0 while the commits failed). The follow-up commits carry the remainder. Net content is identical to a clean run; only the commit boundaries differ.

71 evictions total. **37** were covered by a Procedure step that already taught the content; **34** required folding the bullet's counterintuitive kernel into that step *first*, per the rule at `evolve-skill:132`. Both list formats in the tail were preserved rather than normalized — 16 dash-bullet skills and 8 numbered skills, with numbered survivors renumbered contiguously 1–6. (An earlier revision of this body said 9 and 10; those came from the generating run's own `listFormat` self-reports, which covered only 19 of 24 skills and were wrong. Counted from disk.)

## Why the scope is 24 and not 16

The issue was filed against a 16-skill tail measured with a dash-only counter, which reads 0 for skills that format pitfalls as numbered lists. The corrected count adds the 8-skill blender/3d-printing family. The counter is now committed as `scripts/audit-skill-sections.js` (PR #402) and handles both formats, so this class of undercount cannot recur silently.

## Verification

Every skill went through an adversarial verify agent instructed to hunt specifically for a **lost negation** — the documented failure mode, after PRs #381 and #387 each lost an inoculating negation and needed review-driven fold-backs.

**21 passed, 3 failed, all 3 repaired.** The failures are the interesting part, because they are exactly what the tier exists to catch:

1. **`navigate-dach-bureaucracy` — a fabricated claim.** The fold asserted "a foreign account does not substitute", which appeared in neither the evicted bullet nor anywhere else in the file. An invented fact is worse than a dropped one: it reads as authoritative and nothing downstream contradicts it. The repair replaced it with a specific, checkable claim — in Switzerland a foreign account will not work for CHF salary or LSV+ direct debit.
2. **`select-print-material` — lost TPU kernel.** The eviction dropped the fact that TPU is hygroscopic despite its low print temperatures, and the editing agent's rationale for skipping the fold was inverted by the evidence. Restored in negation form: "a dry box is required, not optional; wet TPU does not fail loudly, it just prints stringy with weak, tear-prone layer bonds".
3. **`prepare-print-model` — lost orientation kernel.** Evicted with `folded: false` although the named covering step taught only part of it. Restored as a concrete worked example in Step 4.

Three further advisory findings from the verify tier were folded in on top:

- **`version-ml-data` — a factual error introduced by an otherwise good fold.** It claimed "a rerun always means something changed", which `dvc repro -f` falsifies. Tightened to keep the negation while naming the exception.
- **`optimize-cloud-costs`** — the evicted quota bullet's operational mitigation (review monthly, communicate before enforcing) survived nowhere; appended to the Step 7 On-failure block.
- **`setup-local-kubernetes`** — the 2-CPU-core prerequisite had landed only in reactive failure-path text; promoted to Inputs where it belongs as a pre-flight requirement.

Corpus checks: all 24 at exactly 6 pitfalls, all six required sections present, all under 500 lines, no CRLF introduced.

## Model provenance — please read before approving

This pass spans two models. A Fable 5 workflow hit the usage-credit ceiling mid-run and it resumed on Opus 4.8, so **no skill here carries Fable verification** — the interruption killed all 24 first-run verify agents.

- **Fable-authored eviction decisions (8):** `assess-github-repo-security`, `deploy-ml-model-serving`, `optimize-cloud-costs`, `orchestrate-ml-pipeline`, `prune-agent-memory`, `setup-local-kubernetes`, `setup-service-mesh`, `version-ml-data`.
- **Fable-authored folds, Opus-authored evictions (4):** `build-feature-store`, `configure-api-gateway`, `enforce-policy-as-code`, `navigate-dach-bureaucracy`. Fable's fold edits were preserved verbatim; only the "which four to remove" decision is Opus's.
- **Fully Opus (12):** the remainder.
- **All 24 verifications:** Opus.

Tracked in **#400** with a per-skill checklist and the recheck procedure, filed at the maintainer's request. Eviction is judgment work, so the provenance is worth keeping rather than assuming parity.

## Known follow-up

The 10 i18n mirrors of each touched skill will report a stale `source_commit` once this lands. `validate:translations` is warn-only, so this does not block. Refreshing ~240 translated files belongs to the existing bulk-refresh issue **#278**, not here.

Closes #382.

### DIFF
--- a/skills/assess-github-repo-security/SKILL.md
+++ b/skills/assess-github-repo-security/SKILL.md
@@ -15,7 +15,7 @@ license: MIT
 allowed-tools: Read Bash Grep
 metadata:
   author: Philipp Thoss
-  version: "1.0"
+  version: "1.1"
   domain: git
   complexity: intermediate
   language: multi
@@ -348,17 +348,6 @@ absence of the control).
   needs `security_events`; rulesets and Actions permissions need admin. A
   403 is "not assessed", not a GAP — recording it as a gap fabricates a
   finding.
-- **A green check that is not REQUIRED is advisory only**: A CI job that
-  runs and passes gates nothing unless its context is in the ruleset's
-  `required_status_checks`. Do not report a running check as a protective
-  control.
-- **Alerts != fixes**: Dependabot **alerts** (`vulnerability-alerts`, 204)
-  only detect; **security updates** (`automated-security-fixes`) open the
-  fix PRs. Enabling one does not enable the other — assess both.
-- **Treating paid-feature absence as a gap on a public repo**: Secret
-  scanning, push protection, CodeQL, and dependency review are free on
-  public repos; GHAS / Secret Protection / Code Security are private/org
-  products. Their absence on a public repo is N/A, not a gap.
 - **Misreading the auto-commit-bot bypass**: The default `GITHUB_TOKEN` /
   `github-actions[bot]` can never be a ruleset bypass actor. If a repo
   auto-commits to a branch that has `required_status_checks` or a
@@ -368,10 +357,6 @@ absence of the control).
   is the repo default; per-job `permissions:` in the workflow YAML can
   raise or drop it for same-repo events. The API cannot show the effective
   per-workflow grant — note that limitation.
-- **Probing SECURITY.md at only one path**: GitHub auto-detects the policy
-  at the repo root, `docs/`, OR `.github/`. Checking only one path and
-  recording a 404 as a GAP fabricates a finding — a repo with SECURITY.md
-  at root is fully compliant. Only record a GAP if all three 404.
 - **Inferring CodeQL from `security_and_analysis`**: that object carries
   `secret_scanning*`, `dependabot_security_updates`, and `advanced_security`
   — but **no** code-scanning field. CodeQL default-setup state comes only
--- a/skills/build-feature-store/SKILL.md
+++ b/skills/build-feature-store/SKILL.md
@@ -11,7 +11,7 @@ license: MIT
 allowed-tools: Read Write Edit Bash Grep Glob
 metadata:
   author: Philipp Thoss
-  version: "1.0"
+  version: "1.1"
   domain: mlops
   complexity: advanced
   language: multi
@@ -219,7 +219,7 @@ def validate_point_in_time_correctness(training_df, entity_df):
 
 **Expected:** Historical features retrieved successfully, entity_df timestamps preserved, no NaN values for materialized features, point-in-time correctness guaranteed (no future data leakage), feature service groups features logically.
 
-**On failure:** Check entity_df has required columns (entity names + event_timestamp), verify feature view names match registry, ensure offline store has data for requested time range, check for timezone mismatches (use UTC), verify entity IDs exist in source data, inspect logs for SQL query errors, validate feature view TTL covers requested time range.
+**On failure:** Check entity_df has required columns (entity names + event_timestamp), verify feature view names match registry, ensure offline store has data for requested time range, check for timezone mismatches (mixed timezones do NOT raise an error — they silently produce incorrect point-in-time joins; use UTC everywhere), verify entity IDs exist in source data, inspect logs for SQL query errors, validate feature view TTL covers requested time range.
 
 ### Step 6: Serve Features for Real-Time Inference
 
@@ -253,7 +253,7 @@ fs = FeatureStore(repo_path=".")
 
 **Expected:** Online features retrieved in <10ms for single entity, batch retrieval scales efficiently, on-demand transformations execute correctly, request-time features merged with batch features, API responds quickly (<50ms end-to-end).
 
-**On failure:** Check online store populated (run materialize if empty), verify Redis/DynamoDB connectivity and latency, ensure entity keys exist in online store, check for cold start issues (warm up cache), verify on-demand transformation logic, monitor online store memory/CPU usage, check network latency between service and online store.
+**On failure:** Check online store populated (run materialize if empty), verify Redis/DynamoDB connectivity and latency, ensure entity keys exist in online store (missing keys return None values, NOT an error — handle them gracefully in serving code), check for cold start issues (warm up cache), verify on-demand transformation logic, monitor online store memory/CPU usage, check network latency between service and online store.
 
 ## Validation
 
@@ -274,13 +274,9 @@ fs = FeatureStore(repo_path=".")
 - **Feature leakage**: Using future data in historical features - always validate point-in-time correctness, use created_timestamp column
 - **Inconsistent transformations**: Different logic for training vs serving - use Feast on-demand views for consistency
 - **Stale features**: Online store not materialized regularly - set up scheduled materialization jobs (cron/Airflow)
-- **Missing entity keys**: Entities in training set not in online store - ensure comprehensive materialization, handle missing keys gracefully
-- **Type mismatches**: Schema types don't match source data - validate dtypes before apply, use explicit Field definitions
 - **Slow online retrieval**: Network latency or overloaded online store - co-locate feature store with inference service, use connection pooling
 - **Large feature views**: Materializing millions of entities is slow - partition by date, use incremental materialization, optimize offline queries
 - **No feature versioning**: Breaking changes affect production models - version feature views, maintain backward compatibility
-- **Timezone confusion**: Mixing timezones causes incorrect joins - always use UTC for timestamps
-- **Ignoring TTL**: Serving expired features - set appropriate TTL values, monitor feature freshness
 
 ## Related Skills
 
--- a/skills/circuit-breaker-pattern/SKILL.md
+++ b/skills/circuit-breaker-pattern/SKILL.md
@@ -13,7 +13,7 @@ license: MIT
 allowed-tools: Read Write Edit Bash Grep Glob
 metadata:
   author: Philipp Thoss
-  version: "1.1"
+  version: "1.2"
   domain: general
   complexity: intermediate
   language: multi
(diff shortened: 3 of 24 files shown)


pjt222:
### Pre-existing defect surfaced and corrected during the repair round

`prepare-print-model` Step 4 contained a physically inverted example, unrelated to the eviction but adjacent to a fold target:

```text
- Orient so layer lines run perpendicular to primary load direction
- Example: Bracket under tension -> print vertically so layers stack along load axis
```

"Layers stack along the load axis" puts every layer interface normal to the tensile load — the delamination-prone orientation the line directly above it tells you to avoid. The example described the weak orientation as if it were the strong one, and the ambiguous noun "layers" (bond plane? stacking direction?) is what let the contradiction sit unnoticed.

FDM parts fail at the layer bond, not through the extruded bead, so a part in tension wants the load carried along continuous beads in the X/Y plane. Step 4 now names the stacking direction (build Z) explicitly and states the mechanism rather than only the rule, so the reader can derive orientation for load cases no table covers.

Filed as **#404** for domain confirmation, since it is a physics claim rather than a documentation one, and the correction is included here rather than left to a follow-up because shipping the eviction while leaving an inverted example next to the fold would be worse than either alone.

pjt222:
### Acceptance criteria, verified

**AC1 — each skill reduced to ≤6 pitfalls.** All 24 at exactly 6, confirmed with the counter shipped in #402, which handles all three list formats in use (dash, numbered, table):

```
$ node scripts/audit-skill-sections.js <24 ids>
… 24 skill(s) reported.   # every row: pitfalls=6, missing sections: -
```

Format preserved rather than normalized — the 10 numbered-list skills stay numbered and renumbered contiguously 1–6; the 9 dash-bullet skills stay dashes. (24 here rather than the issue's 16: the original tail was measured with a dash-only counter that reads 0 for numbered lists, missing the 8-skill blender/3d-printing family.)

**AC2 — every eviction names its covering step, or gives a lowest-value rationale.** 71 evictions across the pass: **37** covered by a Procedure step that already taught the content, **34** required folding the kernel into that step first.

**AC3 — no counterintuitive signal lost.** This is the AC the pass exists to satisfy, given that #381 and #387 each lost an inoculating negation. Every skill went through a verify agent instructed to hunt for a lost negation specifically. 21 passed, 3 failed, all 3 repaired:

- `navigate-dach-bureaucracy` — a **fabricated** claim ("a foreign account does not substitute") present in neither the evicted bullet nor any source. Replaced with a checkable one: in Switzerland a foreign account will not work for CHF salary or LSV+ direct debit.
- `select-print-material` — lost the TPU hygroscopy kernel; restored in negation form ("a dry box is required, not optional; wet TPU does not fail loudly").
- `prepare-print-model` — evicted with `folded: false` although the named step taught only part of it.

Three further advisory findings were folded on top, including a factual error introduced by an otherwise-good fold: `version-ml-data` claimed "a rerun always means something changed", which `dvc repro -f` falsifies.

**AC4 — one PR, per-skill commits.** 24 skill commits (six carry a second commit; see the PR body for why).

### Other gates

`npm run validate:i18n-parity` — 3576 translated skills compared, all `allowed-tools` match. `check-content-style.js --added origin/main` — 0 blocking errors. Required-sections gate — 0 reported. All 9 CI checks green; MERGEABLE/CLEAN.

pjt222:
## Review trail: advocatus-diaboli substitute (Copilot quota-blocked)

Third quota block in a row, so the standing fallback ran again — this time hunting specifically for the #381/#387 failure mode across all 24 skills.

**The headline result is negative, and that is the point: no inoculating negation was lost.** The reviewer went looking for the documented failure and instead found the folds carrying negations explicitly in 13 of the 24 files — `build-feature-store:222` ("mixed timezones do NOT raise an error — they silently produce incorrect point-in-time joins"), `register-ml-model:118` ("MLflow does **not** gate promotion on model quality"), `enforce-policy-as-code:128`, `create-2d-composition:212`, and others. Two evictions it expected to be lost turned out to be already well-homed and correctly needed no fold.

### Folded in — commit `9894436f`

| Severity | Finding | Resolution |
|---|---|---|
| MEDIUM | **Swiss claim false as an absolute.** LSV+ half correct (domestic scheme, foreign IBAN cannot enrol); CHF-salary half wrong — it works over SWIFT, and cross-border commuters are a standing counterexample. The true statement is that most employers refuse the SWIFT fees | Restated with that split, moved to the sub-step where it belongs |
| MEDIUM | **DVC claim incomplete.** "Barring `dvc repro -f`, a rerun always means something changed" misses `always_changed: true` and any stage declaring no deps/outs — so the pitfall caused the debugging cycle it exists to prevent | Both exceptions named |
| MEDIUM | **Gateway TLS eviction had no home.** No `cert-manager`, no expiry guidance anywhere in the Procedure, while Inputs still lists TLS certificates — an outright violation of the eviction rule | Folded into Inputs with the consequence: an expiry is an all-routes outage |
| LOW | **TPU claim contradicted a surviving pitfall four entries below it** ("steam bubbles in the extruder" *is* the loud failure), and invited the inverse inference "no popping, therefore dry" | Restated so absence of symptoms is explicitly not evidence of dryness |
| LOW | **Same kernel folded twice** in `navigate-dach-bureaucracy` (`:268` and `:309`), producing a paragraph that stated its rationale, cross-referenced the sub-step, then restated it | Duplicate removed |
| LOW | **Blender import guidance a non-sequitur.** Relative imports prevent resolving against the wrong top-level module, not circular imports; the fold kept only the half that does not address circularity | Both halves restored |
| LOW | UV clause overstated — a summer outdoors is weak evidence, not "no evidence" | Softened |
| NIT | "Stack layers perpendicular to the neutral axis" reads two ways, one being the failure mode the step warns about | States the orientation directly |

### Worth calling out

The Swiss fold has now been corrected **twice** — first for inventing a claim outright, then for overcorrecting into a different false absolute. A fold that has already been repaired once deserves more scrutiny than an untouched one, not less. Noted in the commit message.

### Both handed-back checks, resolved

The reviewer had no Bash and explicitly flagged two things it could not verify:

- **`create-skill` version step.** Confirmed `1.3 → 1.4` — `git show main:skills/create-skill/SKILL.md` reads `1.3`. One minor step.
- **List-format census.** The reviewer counted 16 dash / 8 numbered from disk and flagged that my PR body said 9 / 10. **The reviewer is right and the PR body was wrong** — those numbers came from the generating run's own `listFormat` self-reports, which covered only 19 of 24 agents and were inaccurate. Body corrected, and the correction is noted in it rather than silently swapped.

`node scripts/audit-skill-sections.js` re-run over all 24 after the fixes: every skill at exactly 6 pitfalls, no missing sections, all under 500 lines.

### Deferred, with reasons

**configure-api-gateway "Resource Limits"** eviction — the reviewer rated this defensible itself (generic Kubernetes hygiene, low rediscovery cost) and did not recommend action.

**setup-service-mesh version-support windows** ("Istio supports n-1 minor versions") — flagged for the author's call rather than as a defect. Keeping the eviction: the surviving Step 1 On-failure line teaches the *check* and names the authoritative source, and pinned version numbers rot fastest of anything in the file.

**The i18n divergence is worse than a stale hash.** The reviewer confirmed `--warn` is genuinely warn-only (`validate-skills.yml:125-126`) and that the blocking parity gate compares `allowed-tools` only, so this PR stays green. But it makes a sharper point than the PR body did: 240 locale files now carry 9–10 pitfalls against a 6-pitfall English source, which is *semantic* drift that the freshness signal understates. Added to #278.

Issue #408, opened 2026-07-23
evolve-skill: consider a 're-repaired fold' scrutiny rule — the one fold corrected twice was wrong both times


pjt222:
## Observation

During the #382 eviction pass (PR #406), one fold was corrected twice and was wrong on both passes:

1. **First state — fabricated.** The fold asserted "a foreign account does not substitute", a claim present in neither the evicted bullet nor anywhere else in the file. Caught by the verify tier.
2. **First repair — overcorrected.** The replacement claimed "in Switzerland a foreign account will not work for CHF salary or LSV+ direct debit at all". The LSV+ half is right (domestic scheme, foreign IBAN cannot enrol); the CHF-salary half is false (it works over SWIFT; cross-border commuters are a standing counterexample). Caught by the PR-level adversarial review.
3. **Second repair** — split into the true claim (LSV+ impossible; most employers refuse the SWIFT fees) and moved to the sub-step where the guidance belongs.

Every other fold in the pass that passed verification on the first attempt has held up. The single fold that needed a repair round is the single fold that was still wrong afterwards.

## Why this might be a rule rather than a coincidence

A repair is authored under different conditions than the original fold: the agent is handed a *problem statement* ("this kernel was lost", "this claim is unsupported") rather than the source material, and the natural failure is to satisfy the stated complaint as directly as possible. Replacing a fabricated claim with a confident, specific-sounding claim does satisfy "stop fabricating" — while introducing a fresh factual error that the original complaint said nothing about.

If that mechanism is real, repaired folds carry *higher* residual risk than untouched ones, and current practice has it backwards: the repair round is the last thing that happens before the diff is considered settled.

## Proposal (for discussion, not yet actioned)

Consider adding to `evolve-skill` near the eviction rule at `:132`: when a fold is repaired in response to a verification failure, re-verify the repaired text against the *source material*, not against the problem statement — a repair that satisfies the complaint can still introduce a new false claim.

## Evidence bar

Deliberately filed as an issue rather than shipped as a skill edit. The fold-first rule at `evolve-skill:132` was encoded only after the same failure recurred **independently twice** (PRs #381 and #387). This is one instance. It should earn its place the same way, or be dropped if the next eviction pass shows repaired folds holding up fine.

## Acceptance criteria

- [ ] Decision recorded on whether one instance is enough, or whether to wait for a second
- [ ] If adopted, the rule names what to re-verify against (source material, not the complaint)
[...]

Issue #405, opened 2026-07-23
evolve-skill: Step 4.5's source_commit bump hides real staleness; 10 locale copies teach the superseded semver table


pjt222:
### 1. Ten locale copies of `evolve-skill` carry the superseded semver table

PR #402 corrected `evolve-skill` Step 5, whose table mapped an additive change to `Minor: 1.0 → 2.0` (a major bump with a minor label) and called `1.0 → 1.1` a "Patch" bump though skill versions are two-part. The English source and `guides/creating-skills.md` are fixed; the translations still teach the wrong rule:

```
i18n/de/skills/evolve-skill/SKILL.md
i18n/zh-CN/skills/evolve-skill/SKILL.md
i18n/ja/skills/evolve-skill/SKILL.md
i18n/es/skills/evolve-skill/SKILL.md
i18n/caveman{,-lite,-ultra}/skills/evolve-skill/SKILL.md
i18n/wenyan{,-lite,-ultra}/skills/evolve-skill/SKILL.md
```

The version numbers in that table are language-independent, so this is a factual error sitting in ten files, not a translation-quality question.

### 2. `evolve-skill` Step 4.5 prescribes a `source_commit` bump that hides staleness

Step 4.5 is marked **"Required when translations exist… Do not skip"** and its stated rationale is that stale `source_commit` values "cause `npm run validate:translations` to report false staleness warnings across all locales."

That rationale only holds when the translation was actually re-translated. When the English source changes and the locale files do not, the staleness warning is **true, not false** — and bumping `source_commit` to the new English commit marks un-retranslated content as fresh, suppressing the one signal that would have surfaced it. `check-translation-freshness.js` derives staleness from `git log <source_commit>..HEAD`, so a bumped commit makes the drift invisible.

Step 4.5 currently reads as four sequential steps (bump, flag in commit message, regenerate status), so following it literally on any change that is not accompanied by re-translation actively degrades the freshness signal.

PR #402 therefore deviated deliberately: it flagged the locales in the commit message (Step 4.5's step 3) and did **not** bump `source_commit`. That deviation should either become the documented rule or be replaced with something better.

## Proposal

- Fix the table in the ten locale copies (numbers and labels only — no re-translation needed).
- Rewrite Step 4.5 to distinguish the two cases explicitly:
  - **Re-translated in this change** → bump `source_commit`, it is now accurate.
[...]

PR #409, opened 2026-07-23
evolve(create-skill): add a pitfall on unverified tool-behavior claims

pjt222:
## Why

The sharpest lesson from today's sweep, encoded so it survives the session.

Three Helm pitfalls I authored in #402 were factually wrong. None was written carelessly — each was written **confidently**, and that was the mechanism:

- "hook Jobs accumulate and a re-install fails on the name collision" — Helm's default `before-hook-creation` policy deletes the prior hook, so no collision occurs
- "an empty tag makes each install pull whatever `latest` points to" — with the skill's own `| default .Chart.AppVersion` template it resolves to AppVersion, and a bare trailing colon is `InvalidImageName`, not `latest`
- "values deep-merge, they do not replace" — true for maps, false for lists, and contradicted by the skill's own `values-dev.yaml`

The adversarial reviewer caught all three, but only because it checked upstream docs instead of accepting the prose. Nothing downstream contradicts a confidently-wrong claim, so it survives review by default and misleads every run that follows.

The generalizable rule: **confidence is the signal to verify, not to skip verifying** — and an observed failure mode beats a believed one.

## The eviction

`create-skill` was at the 6-pitfall cap, so per the rule at `evolve-skill:132` this evicts rather than appends. Removed **Description too long**, the lowest-rediscovery-cost entry in the set — every claim it made is already taught three times:

| Where | What it already says |
|---|---|
| Frontmatter template `:74-77` | "Must be clear enough for an agent to decide whether to activate this skill from the description alone. Max 1024 characters. Start with a verb." |
| Step 2 Expected `:99` | repeats the 1024 limit and activation triggers |
| Validation checklist `:365` | `description` is under 1024 characters |

Its one distinct phrase — "front-load the key information" — is what "Start with a verb" already instructs. **No fold was needed**, because no kernel was unique to it.

### Considered and rejected as eviction candidates

- **Stale cross-references** looks redundant with Step 9's On-failure block, but is not. Step 9 teaches the *outbound* direction (verify my references resolve); the pitfall teaches the *inbound* one (renaming a skill breaks references held by other skills). Evicting it would have required a fold first.
- **Missing On failure** is taught at `:186`, but its "agents can't improvise" clause is the motivating *why*, stated nowhere else.

## Verification

`node scripts/audit-skill-sections.js create-skill` → 6 pitfalls, all six sections, 411 lines. Corpus gate, `validate:i18n-parity` (3576 files), and `check-content-style --added` all clean. LF only.

Version `1.4 → 1.5` — additive under the criterion clarified in #388: a caller following 1.4 still gets a correct result.

## Translations

Flagged for re-sync (10 locales): de, zh-CN, ja, es, caveman{,-lite,-ultra}, wenyan{,-lite,-ultra}. Changed section: Common Pitfalls.

Per **#405**, `source_commit` is deliberately **not** bumped — the locale copies were not re-translated, and bumping would mark stale content as fresh and suppress the very freshness signal that should surface it.

## Related

Sibling of **#408**, which records the adjacent observation (a repaired fold carries higher residual risk than an untouched one) and is deliberately *not* yet encoded — one data point, where the fold-first rule earned its place only after two independent recurrences. This pitfall has three.

### DIFF
--- a/skills/create-skill/SKILL.md
+++ b/skills/create-skill/SKILL.md
@@ -379,9 +379,9 @@ npm run translation:status
 - **Missing On failure**: Every step that can fail needs recovery guidance. Agents can't improvise — they need the fallback spelled out.
 - **Overly broad scope**: A skill that tries to cover "Set up entire development environment" should be 3-5 focused skills instead. One skill = one procedure.
 - **Stale cross-references**: When renaming or removing skills, grep for the old name in all Related Skills sections.
-- **Asserting tool behaviour you have not checked**: The claim you feel surest about is the one to verify. "The default is X", "this fails with Y", "the tag falls back to latest" — such statements read as authoritative and nothing downstream contradicts them, so a wrong one survives review and misleads every run that follows. Check the primary docs while writing, and prefer a failure you have observed over one you believe exists.
 - **Authoring at 500-line limit for single language**: An English skill at 490 lines will exceed 500 when translated to German (~10-20% expansion) or CJK languages. Target ~400 lines for the English source and use progressive disclosure (`references/EXAMPLES.md`) for the rest.
 - **Avoid `git mv` on NTFS-mounted paths (WSL)**: On `/mnt/` paths, `git mv` for directories can create broken permissions (`d?????????`). Use `mkdir -p` + copy files + `git rm` the old path instead. See the [environment guide](../../guides/setting-up-your-environment.md) troubleshooting section.
+- **Asserting tool behavior you have not checked**: The claim you feel surest about is the one to verify — "the default is X" reads as authoritative, so a wrong one survives review by default. Cite the doc you checked, and prefer a failure you observed over one you believe exists.
 
 ## Examples
 
--- a/skills/create-skill/SKILL.md
+++ b/skills/create-skill/SKILL.md
@@ -13,7 +13,7 @@ license: MIT
 allowed-tools: Read Write Edit Bash Grep Glob
 metadata:
   author: Philipp Thoss
-  version: "1.4"
+  version: "1.5"
   domain: general
   complexity: intermediate
   language: multi
@@ -379,7 +379,7 @@ npm run translation:status
 - **Missing On failure**: Every step that can fail needs recovery guidance. Agents can't improvise — they need the fallback spelled out.
 - **Overly broad scope**: A skill that tries to cover "Set up entire development environment" should be 3-5 focused skills instead. One skill = one procedure.
 - **Stale cross-references**: When renaming or removing skills, grep for the old name in all Related Skills sections.
-- **Description too long**: The description field is what agents read to decide activation. Keep it under 1024 characters and front-load the key information.
+- **Asserting tool behaviour you have not checked**: The claim you feel surest about is the one to verify. "The default is X", "this fails with Y", "the tag falls back to latest" — such statements read as authoritative and nothing downstream contradicts them, so a wrong one survives review and misleads every run that follows. Check the primary docs while writing, and prefer a failure you have observed over one you believe exists.
 - **Authoring at 500-line limit for single language**: An English skill at 490 lines will exceed 500 when translated to German (~10-20% expansion) or CJK languages. Target ~400 lines for the English source and use progressive disclosure (`references/EXAMPLES.md`) for the rest.
 - **Avoid `git mv` on NTFS-mounted paths (WSL)**: On `/mnt/` paths, `git mv` for directories can create broken permissions (`d?????????`). Use `mkdir -p` + copy files + `git rm` the old path instead. See the [environment guide](../../guides/setting-up-your-environment.md) troubleshooting section.
 


pjt222:
## Review trail: adversarial substitute (Copilot quota-blocked, 4th consecutive)

The review was asked to test one thing above all: whether a pitfall *about* unverified claims manages to avoid committing its own sin. **It did not**, and the finding is the most useful thing in this PR.

### Folded in — commit `8d0c0b66`

| # | Severity | Finding |
|---|---|---|
| F1 | MEDIUM | The text claimed "a wrong one survives review and misleads every run that follows" — an absolute contradicted by this pitfall's own provenance, since review caught all three wrong Helm claims. The PR body had the qualified form ("by default"); the SKILL.md text had dropped it. An author reading the absolute would conclude review is useless against this class, inverting the lesson. Restored. |
| F2 | LOW | The example "the tag falls back to latest" is **true** in its likeliest referent — an image reference with no tag does default to `latest`. The originating text said "an *empty* tag"; the qualifier was lost in compression, so the one concrete example of a *suspect* claim was a correct one. Dropped rather than patched: "the default is X" carries a variable and cannot be misread as an assertion. |
| F8 | NIT | "Check the primary docs while writing" was the only fix in the set naming no artifact, against the bar this skill sets at `:222`. Now "cite the doc you checked" — verifiable in review. Also trimmed 74 → 51 words (siblings run 18–41). |
| F6 | NIT | Moved to the end of the list per the placement rule at `evolve-skill:132`; it had been substituted in place at position 4. |
| F7 | NIT | "behaviour" → "behavior"; corpus is 342 en-US to 11 en-GB. |

### Rationale corrections (F3, F4, F5)

These do not change the outcome, but eviction rationales are the precedent record for the next eviction, and an inaccurate redundancy table is precisely how a kernel gets lost. Correcting them here rather than leaving the record wrong.

**F3 — my redundancy equivalence was wrong; the decision survives for a different reason.** I claimed "front-load the key information" is what "Start with a verb" already instructs. It is not — "start with a verb" constrains the first token, "front-load" constrains ordering by importance, and an author can satisfy one while violating the other. The eviction still clears the bar, because `evolve-skill:132` requires a fold only for a **counterintuitive fact**, and front-loading is style advice whose underlying fact — *the description is read at discovery, before the body is loaded* — is stated verbatim at **`SKILL.md:131`**, which I never cited. The reviewer found the citation I should have used. Practical loss is near zero; `agents/skill-reviewer.md:112` also retains "trim to front-load key info".

**F4 — I assessed 2 of 5 survivors and skipped the strongest alternative.** "Authoring at 500-line limit" has *no* unique kernel: the ~400-line target, the 10–20% German expansion, the CJK note and the >500 consequence are all at `:175-178`, with checklist items at `:369` and `:371` and the progressive-disclosure remedy in Step 12. It needed no fold, where the entry I evicted needed the `:131` argument to clear the bar. My choice remains defensible on the rule's own terms — `evolve-skill:132` ranks "by what they save", and a translation-expansion trap saves more than a character cap stated inline where you type it — but the body should have said so instead of leaving the strongest alternative unmentioned.

**F5 — right conclusion, wrong route.** I rejected "Missing On failure" because its "agents can't improvise" clause is unique (verified: sole occurrence repo-wide). Uniqueness is not a veto — `evolve-skill:132` provides fold-then-evict for exactly that case, and I applied that framing correctly to "Stale cross-references" one sentence earlier, so the treatment was internally inconsistent. The correct rejection is that it would require a fold *and* its saving is higher than the evicted entry's.

### Confirmed clean

The "Stale cross-references" rejection **survives** verification — `:240` is strictly outbound ("verify each referenced skill exists"), the pitfall at `:380` is inbound ("grep for the old name in all Related Skills sections"). Genuinely distinct; evicting it would have needed a fold. The new pitfall is **not** redundant: nearest neighbours are `:177` and `:222`, and the latter is scoped to *sourcing* pitfalls, not to verifying claims in Expected/On-failure prose. Diff scope clean at 2 lines, 1 file. `npm run translation:status` regenerates to a no-op.

Two points in the PR's favour that I had not made: this eviction **continues** the ranking from `920f9a94` (#406), which removed the two lowest-value entries from this same list and left "Description too long" as next-lowest — so it is a continuation, not a reversal. And the translation deviation is not a deviation from `evolve-skill` as a whole: its own pitfall list sanctions the disjunction ("update `source_commit` **or** flag them for re-translation in the commit message"), and the commit message flags all ten locales and the changed section.

### Translation deviation — verified defensible

`check-translation-freshness.js` is genuinely warn-only (`--warn` in the workflow; the script exits 1 only when `!WARN_ONLY` at `:152-153`). More decisive: the locale copies were **already stale before this PR** — `source_commit` is `496be0a2` / `82c77053` while `create-skill` has been modified five times since. Bumping would stamp five commits of untranslated drift as fresh, which is exactly the falsification #405 describes. #406 set the same precedent.

Residual LOW, acknowledged: Step 4.5's Expected block is knowingly unmet while #405 is open. The clean sequencing is to land the Step 4.5 rewrite first so this stops needing per-PR justification.

### Filed separately

`evolve-skill`'s "Stale translations after evolution" pitfall still cites **"1,288 translation files"** and **"up to 4 locale files"**; the corpus is 10 locales and 3,576 skill files by the parity gate's own count. Out of scope here, filed per the standing order.

Issue #457, opened 2026-07-29
cli: universal detection rule is `existsSync(...) || true`, so the check is dead and universal is always detected


pjt222:
## Summary

`cli/lib/detector.js:20`:

```js
{ id: 'universal', displayName: 'Universal (.agents/)', check: (dir) => existsSync(resolve(dir, '.agents')) || true, marker: '.agents/', scope: 'project' },
```

`existsSync(...) || true` is unconditionally `true`. The `existsSync` call is dead code, and `universal` is detected in every directory regardless of whether `.agents/` exists.

## Consequences

**1. `detect` reports a framework that is not there.** In an empty directory:

```
$ cd /tmp/empty && node cli/index.js audit --source <repo>

Audit results:

Universal (.agents/):
  WARN No skills installed in .agents/skills/
```

There is no `.agents/` in that directory at all. The audit reports on a framework the project does not use.

**2. `detectFrameworks()` can never return an empty list**, so downstream code guarding the empty case is unreachable. This surfaced while implementing #439: the "no frameworks detected" warning added there cannot fire, and the guard is retained with a comment pointing here rather than being silently dropped.

**3. The intent is ambiguous from the code.** Two readings, and the fix differs:

- *Intentional* — `.agents/` is the cross-client standard and `universal` should always be offered as an install target. Then the rule should say so plainly: `check: () => true`, with a comment. The current form reads like a condition when it is a constant.
- *Accidental* — someone forced it on (to test, or to make `install` default to universal) and left the `|| true`. Then the fix is to drop `|| true` and let it detect honestly.

The `|| true` form is indistinguishable from a debugging leftover, which is why this is worth resolving either way rather than leaving as-is.

## Suggested fix

Decide which reading is correct, then make the code state it unambiguously:

```js
// Universal is the cross-client fallback: always an eligible target,
// whether or not .agents/ exists yet.
{ id: 'universal', ..., check: () => true, ... }
```

or

```js
{ id: 'universal', ..., check: (dir) => existsSync(resolve(dir, '.agents')), ... }
```

If the second is chosen, check that `install` still has a sensible default target in a fresh project, since that is the most likely reason the `|| true` was added.

## Acceptance criteria

- [ ] The `universal` check is either a plain constant with a comment, or a real filesystem test — no `x || true`
- [ ] If it becomes a real test: `detect` in an empty dir no longer lists Universal, and `audit` there emits the "no frameworks detected" warning (which becomes reachable)
[...]

Issue #456, opened 2026-07-29
cli: --version reports 0.1.0 against a 1.3.0 package, and cli/README advertises Node 18+ against engines >=22.12.0


pjt222:
## Summary

Two independent pieces of the CLI's self-description have drifted from `package.json`. Both are one-line fixes; the first is user-visible on the published npm package.

## 1. `--version` reports 0.1.0 while the package is 1.3.0

`cli/index.js:38`:

```js
  .version('0.1.0')
```

against `package.json` `"version": "1.3.0"`. Measured:

```
$ node cli/index.js --version
0.1.0
```

This ships. The package is published to npm as `agent-almanac@1.3.0` with two bins (`agent-almanac`, `almanac`), so every user who runs `--version` — including anyone answering "what version are you on?" in a bug report — gets the wrong answer, and has no way to notice.

It also means the version has been wrong for every release since 0.1.0; the string is hardcoded and nothing reconciles it.

**Fix:** read it from `package.json` rather than hardcoding, so it cannot drift again:

```js
import pkg from '../package.json' with { type: 'json' };
// ...
  .version(pkg.version)
```

(Import attributes need Node 20.10+; the engines floor is already `>=22.12.0`, so this is safe. If the JSON import is undesirable, `createRequire` works too — but a hardcoded string that must be manually synced is what produced this bug.)

## 2. `cli/README.md` advertises a Node floor three majors below the real one

`cli/README.md:47`:

```
Requires Node.js 18+.
```

against `package.json:29`:

```json
  "engines": {
    "node": ">=22.12.0"
  },
```

A user on Node 18 or 20 follows the README, installs, and gets an npm `EBADENGINE` warning — or a runtime failure on any syntax the CLI uses that predates their runtime. `engines` is advisory (npm warns rather than blocks), so the README is actively steering people into an unsupported configuration rather than being caught by the installer.

**Fix:** update `cli/README.md:47` to match `package.json:29`. Worth a grep for the same claim elsewhere in the docs and its i18n mirrors.

## Acceptance criteria

- [ ] `node cli/index.js --version` matches `package.json` version
- [ ] A test asserts the two agree, so the next release cannot re-drift (the existing `shows version` test at `cli/test/cli.test.js:571-574` only asserts `/\d+\.\d+\.\d+/`, which `0.1.0` satisfies — it is why this was never caught)
- [ ] `cli/README.md` states the same Node floor as `package.json` `engines`

## Context

[...]

Issue #455, opened 2026-07-29
cli: the chalk fallback proxy is broken by construction — chalk.hex() returns a string, so degradation crashes instead of dropping colour


pjt222:
## Summary

Both chalk import sites install a "no colours" fallback for when chalk is unavailable. The fallback is broken by construction: it satisfies the direct style functions but not the *factory* functions (`hex`, `bgHex`, `rgb`), which must return a function and instead return a string. If chalk ever fails to import, the CLI does not degrade to plain text — it crashes with `TypeError`.

## The bug

`cli/lib/reporter.js:8-15`:

```js
let chalk;
try {
  chalk = (await import('chalk')).default;
} catch {
  // Fallback: no colors
  const identity = (s) => s;
  chalk = new Proxy({}, { get: () => identity });
}
```

`cli/lib/tui.js:13-19` is the same pattern with `id` instead of `identity`.

Every property access returns `identity`. That is correct for direct styles and wrong for factories:

```
chalk.dim("x")            -> "x"                     ✓ works
chalk.hex("#FF6B35")      -> "#FF6B35"  (a string)   ✗ should be a function
chalk.hex("#FF6B35")("f") -> TypeError: c is not a function
```

Reproduced directly:

```js
const identity = (s) => s;
const chalk = new Proxy({}, { get: () => identity });
chalk.hex('#FF6B35')('flame');   // TypeError: c is not a function
```

## Blast radius

The palettes are built at module load from `chalk.hex(...)`, so the failure is immediate and total, not a degraded edge case:

- `cli/lib/campfire-reporter.js:24-28` — 5 × `chalk.hex`
- `cli/lib/tui.js:23-27` — 5 × `chalk.hex`
- `cli/lib/pixel-renderer.js:21,26` — `chalk.hex(hex)` and `chalk.bgHex(hex)`, cached per colour

With `import('chalk')` forced to throw, **16 tests fail**: `TypeError: C.flame is not a function` (`campfire-reporter.js:160`), `C.warm is not a function` (`campfire-reporter.js:126,284`), `fgColor(...) is not a function` (`pixel-renderer.js:76`).

## Severity

[...]

Issue #454, opened 2026-07-29
test(cli): the suite is structurally blind to chalk output — passes byte-identically with chalk fully broken


pjt222:
## Summary

`cli-test` cannot observe chalk's output at all. It passes identically whether chalk works, changes its rendering, or is entirely broken. Any regression in the CLI's colour output — the campfire palette, the half-block pixel art, the audit markers — is invisible to CI and will stay invisible.

Surfaced while verifying the chalk 5→6 major bump (#449). That PR is genuinely safe, but its green `cli-test` was **not** the evidence that established it; a separate empirical test was needed. This is the fourth instance of the repo's documented false-green pattern (#363, #365, #373, #443).

## Why the suite is blind

**Colour is off in tests, and nothing turns it on.** `rg -n "FORCE_COLOR|NO_COLOR"` across `cli/`, `package.json`, and `.github/workflows/ci-node-cli.yml` returns exactly one hit — a prose comment at `cli/test/cli.test.js:48`. Nothing sets either variable.

`run()` at `cli/test/cli.test.js:35-37` passes no `env:`:

```js
function run(args) {
  return execSync(`${CLI} ${args}`, { cwd: ROOT, encoding: 'utf8', timeout: 10000 });
}
```

Capturing stdout makes the child's stdout a pipe rather than a TTY, so `chalk.level` is 0 and the CLI emits no ANSI at all. Measured: `node cli/index.js {list --domains, detect, audit, campfire --all}` piped → ESC bytes = 0, 0, 0, 0.

**Nothing asserts on colour.** `cli/test/cli.test.js:56` is the sole ANSI-aware line in the 843-line file, and it *strips*:

```js
  const lines = out.replace(/\x1b\[[0-9;]*m/g, '').split('\n');
```

To be clear, that strip is **correct** for what it does — it is the #373 fix, and the comment at `:47-49` shows the author knew chalk colours the header. The defect is that nothing complements it: there is not one assertion on an escape sequence anywhere in the file.

**The one colour-dependent branch is pinned to its off state.** `cli/test/cli.test.js:673-674`:

```js
  it('canRenderPixelArt returns false in non-TTY test env', () => {
    assert.equal(canRenderPixelArt(), false);
```

`canRenderPixelArt()` (`cli/lib/pixel-renderer.js:34`) is the only consumer of `chalk.level`, and the suite asserts it is disabled.

## Byte-level proof

[...]

PR #458, opened 2026-07-29
feat(cli): give audit a real exit code, and mark crashes structurally (#439)

### LINKED ISSUE #439: cli: `audit` exits 0 when an adapter reports errors, so a broken audit is invisible to automation
## Summary

`agent-almanac audit` always exits 0, whatever it finds. `auditAll()` catches any adapter throw and converts it into a result entry:

```js
// cli/lib/installer.js:129-136
} catch (err) {
  results.push({
    framework: adapter.constructor.displayName,
    ok: [], warnings: [],
    errors: [`Audit failed: ${err.message}`],
  });
}
```

`cli/index.js:280-283` then prints the results and returns. Nothing inspects `errors` to set an exit code.

## Why it matters

This is the mechanism that let #365 ship and survive. `ClaudeCodeAdapter.audit()` threw `ReferenceError: base is not defined` on every invocation; the throw was converted to a printed `ERR Audit failed: base is not defined` line and the process still exited 0. Any script or CI step doing `agent-almanac audit && ...` saw success. The printed output was the only signal, which is why #373 had to be about assertion strength rather than exit status.

Reproduced on `main` before the fix:

```
$ node cli/index.js audit; echo "exit=$?"
Claude Code:
  ERR  Audit failed: base is not defined
exit=0
```

## Proposal

Exit non-zero when any adapter result has a non-empty `errors` array. Warnings stay exit 0.

This is a behaviour change to a shipped CLI command, so it is deliberately **not** bundled into the #365/#373 fix. Points to settle first:

- Semantics: is a broken symlink (a real finding) an error exit, or only an adapter *crash*? These are currently merged into the same `errors` array by two very different paths.
- Whether a `--strict` / `--exit-code` flag is preferable to changing the default, given anyone already scripting against `audit`.
- Whether the crash path should stay swallowed at all, or surface as a distinct exit code from ordinary findings.

## Acceptance criteria

- [ ] Decision recorded on default-change vs opt-in flag.
- [ ] `audit` signals adapter failure through its exit status under the chosen semantics.
- [ ] A test asserts the exit code for: clean audit, findings-present audit, and adapter-throws audit.
- [ ] Interaction with #438 considered — that issue makes three more adapters capable of producing errors.

pjt222:
Closes #439.

`audit` printed `ERR` lines and exited 0, so a broken install was invisible to any automation that checked status. That is how #365 survived three weeks, and why B11 had to grep stdout to say anything at all (#443).

Staged as two commits, because the second is not honestly reviewable without the first.

## 1. Mark crashes structurally (`2989f705`)

`auditAll()` turned an adapter throw into an `errors[]` entry with the same shape as a real finding — same keys, same types, same framework name. The only discriminator was that a crash's message starts with `Audit failed: `, i.e. a stdout grep relocated into the data structure. That is the fragility `scripts/validate-integrity.sh:528` warns against.

The distinction is not cosmetic. A finding means the adapter ran and reported something; a crash means it produced *no verdict at all*, so `ok[]` is empty and a wholly broken install audits as "nothing installed, nothing wrong".

`auditAll()` now sets `crashed` on every entry and carries the original throw as `error`. Normalised in `auditAll` rather than across all 14 adapters: returning from `audit()` is what proves the adapter ran to completion, so that is the only place able to set it authoritatively. No behaviour change — `printAudit()` reads only `ok`/`warnings`/`errors`, so output bytes and exit codes are untouched.

## 2. Give audit a real exit code (`30bfe69e`)

| Code | Meaning |
| --- | --- |
| `0` | Audit ran and found no errors |
| `2` | An adapter crashed, so that framework has no verdict |
| `3` | Audit ran and found errors |

**Why not 1.** It is already taken three times over: `getContext()` exits 1 for an undetectable almanac root and for an unknown `--framework`, and node exits 1 with `ERR_MODULE_NOT_FOUND` when deps are missing. Reusing it would leave automation unable to tell *"the CLI is not installed here"* from *"your install is broken"* — precisely the ambiguity B11 died of.

**Crash outranks finding**, since a crash is the more urgent report. **Warnings never affect the code** — they describe ordinary states like `No Copilot skills installed`, so failing on them would make a machine with one framework installed exit non-zero for every other adapter. The full report prints *before* the exit code is set, so a non-zero status never costs the user the reason for it.

**Harness.** Adds `runAllowFail()` and switches the `audit` `before()` block to it. `execSync` throws on non-zero and leaves stdout on `err.stdout`, so without this the first real finding would take down the whole audit block with an opaque error instead of a legible failure. `run()` itself stays throwing — the gather and scatter tests assert on that.

## Verification

88 → 101 tests, all green.

Both commits were checked by breaking the subject and watching the gate go red, per the discipline from #443:

- **Commit 1:** removing the `crashed` flag from `installer.js` fails exactly the 4 new tests and leaves the other 88 green — so they are specific to the subject, with no collateral coupling.
- **Commit 2:** verified end to end rather than only by unit test. A fixture directory with a dangling `.github/skills` symlink exits **3** (`ERR 1 broken skill symlinks`); the *same fixture* with every link valid exits **0**. The dangling link is the only difference between red and green.

One test pins the numeric codes literally rather than via `AUDIT_EXIT`, since asserting only through the constant would let a renumbering pass silently.

## One caveat, stated rather than buried

The "no frameworks detected" warning added here is **currently unreachable**. The universal rule at `cli/lib/detector.js:20` is `existsSync(...) || true`, which always matches, so detection never returns an empty list. The guard is three lines and a silent false-clean to omit, so it is kept — with a comment at the call site saying exactly this, and filed as #457. Please do not read its presence as evidence the empty case has been observed.

## Deliberately not in scope

Wiring the CI gate. `.github/workflows/ci-node-cli.yml` is the right home for it (it runs `npm ci` and triggers on `cli/**`, the two structural reasons B11 could never fire) — but adding a step that can turn `main` red is its own decision, and this PR is the enabling change. Happy to follow up.

Also out of scope, found along the way and filed separately: #453 (campfire tests mutate the real repo root), #454 (suite is blind to chalk output), #455 (chalk fallback proxy is broken by construction), #456 (`--version` drift), #457 (the detector `|| true`).

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -32,6 +32,31 @@ agent-almanac audit                  # Health check installed content
 agent-almanac uninstall <names...>   # Remove installed content
 ```
 
+### Audit exit codes
+
+`audit` always prints its full report before exiting, so a non-zero status never
+costs you the reason for it.
+
+| Code | Meaning |
+| --- | --- |
+| `0` | Audit ran and found no errors |
+| `2` | An adapter crashed, so that framework has no verdict at all |
+| `3` | Audit ran and found errors |
+
+`1` is deliberately not used for audit findings. It already means "usage or
+loader error" — an unknown `--framework`, an undetectable almanac root, or
+missing dependencies — so reusing it would leave automation unable to tell *"the
+CLI is not installed here"* from *"your install is broken"*.
+
+A crash outranks a finding: a finding is a completed audit that found something,
+whereas a crash means that framework produced no verdict at all. Warnings never
+affect the exit code, since they describe ordinary states such as `No Copilot
+skills installed`.
+
+```bash
+agent-almanac audit || echo "audit reported problems (exit $?)"
+```
+
 ## Install
 
 ```bash
--- a/cli/adapters/base.js
+++ b/cli/adapters/base.js
@@ -18,6 +18,10 @@
  * @property {string[]} ok - OK messages
  * @property {string[]} warnings - Warning messages
  * @property {string[]} errors - Error messages
+ * @property {boolean} [crashed] - True when audit() threw and produced no verdict.
+ *   Set by auditAll(), not by adapters — see cli/lib/installer.js. Absent on
+ *   entries obtained by calling an adapter's audit() directly.
+ * @property {Error} [error] - The original throw, present only when crashed
  */
 
 export class FrameworkAdapter {
--- a/cli/index.js
+++ b/cli/index.js
@@ -24,7 +24,7 @@ import { loadRegistries, resolveItems, filterSkills, search, findTeam, findAgent
 import { detectAlmanacRoot, resolveTargetDir } from './lib/resolver.js';
 import { detectFrameworks } from './lib/detector.js';
 import { getAdapter, getAdaptersForDetections, listAdapters } from './adapters/index.js';
-import { installAll, uninstallAll, auditAll } from './lib/installer.js';
+import { installAll, uninstallAll, auditAll, auditExitCode } from './lib/installer.js';
 import { loadManifest, resolveManifest, generateManifest, writeManifest } from './lib/manifest.js';
 import { loadState, saveState, recordGather, recordScatter, recordWarm, markWelcomed, getFireStates, getFireState, findSharedSkills } from './lib/state.js';
 import * as campfire from './lib/campfire-reporter.js';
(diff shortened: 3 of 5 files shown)


pjt222:
## Adversarial review round (`90da2bf7`)

Copilot was quota-blocked again (`Copilot was unable to review this pull request because the user who requested the review has reached their quota limit` — `COMMENTED`, zero line comments), so this went through the substitute path: four independent review lenses (correctness, contract/compat, "can this test fail", conventions/prose-accuracy), each finding then handed to a separate agent instructed to refute it and to default to refuted when uncertain.

**8 findings raised, 2 survived refutation.** Both were real, both reproduced, both are now fixed.

### 1. The fix was not covered (major)

Deleting `process.exitCode = auditExitCode(results)` from `cli/index.js` — the entirety of `30bfe69e` — left **all 101 tests passing**.

The only test observing a real process status asserted `0` on a clean repo. But `0` is exactly what the *unfixed* CLI returned; #439 is literally "audit always exits 0, whatever it finds". So the assertion was satisfied equally by the fix and by its absence, and the `auditExitCode` tests cover the reduction without ever seeing whether the command calls it.

This is the same defect class the PR exists to close, reproduced inside the PR. Worth stating plainly: the end-to-end verification in the original PR description was real, but it was performed **by hand** and therefore lived nowhere CI could re-run it.

Replaced with three tests that spawn the real CLI:

| Fixture | Exit |
| --- | --- |
| every link resolves | `0` |
| clean fixture **+ one dangling link** | `3` |
| a file where the adapter expects a directory (`readdirSync` → `ENOTDIR`) | `2` |

The findings fixture is the clean fixture plus one dangling link, so that link is the only difference between green and red. The crash is a genuine adapter throw, not a synthesised one.

Mutation-verified: with the assignment removed, `findings` and `crashed` both fail. The same mutation previously failed nothing.

### 2. The assertion coupled the suite to the developer's `$HOME` (minor)

`detectFrameworks()` appends the global rules unconditionally, so `openclaw` and `hermes` join the adapter list whenever `~/.openclaw` or `~/.hermes` exists — and their `audit()` reads `homedir()`, not the project. Since the exit code reduces over *every* adapter, one stale symlink in either would turn this repo's suite red on an untouched checkout, and block `npm publish` through `prepublishOnly`.

The new tests redirect `HOME` at a fixture, matching the existing broken-symlink block. Verified with a negative control: a `HOME` containing a dangling `~/.openclaw/workspace` link makes a bare `audit` exit `3`, while the suite stays **103/103** green.

Fixtures also moved to the OS temp dir via `mkdtempSync` instead of under `ROOT`, so a crashed run cannot poison the next one (cf. #453).

### Refuted, for the record

Six findings did not survive. Two are worth noting because they targeted prose claims, and prose is this repo's documented blind spot — both were checked independently rather than taken on the refuter's word:

- *"all 14 adapters" — the CLI registers 13.* Refuted correctly. There are 13 concrete adapters **plus** `base.js`, whose `audit()` is a real returning implementation that would need its own `crashed: false`; a per-adapter normalisation skipping it would leave exactly that hole. `2989f705` did edit `base.js`. 14 is the right count for edit sites.
- *Exit-code-1 ambiguity is misattributed as why B11 died.* Refuted: #443's own summary contains the `ERR_MODULE_NOT_FOUND` → `echo $?` → `1` transcript, and records that the tightened B11 detected `exit 1` and aborted the job.

Others refuted: a claimed `runAllowFail` timeout hazard, a CHANGELOG/semver objection, a `skills/design-cli-output` contradiction, and a `prepublishOnly` concern superseded by fix 2.

**Test count across the PR: 88 → 103.**

Issue #460, opened 2026-07-29
skills: no skill covers proving a gate can fail, the discipline behind #363/#365/#373/#443/#458


pjt222:
## Gap

The most load-bearing discipline in this repo's recent history has no skill.

"A green gate is evidence about the gate, not the subject — break the thing it guards and watch it go red before trusting it" is the lesson behind #363, #365, #373, #443, #454, and #458. It is applied by hand every time, and it has been gotten wrong *by someone who was explicitly applying it*: in PR #458 the exit-code fix was verified end to end manually (dangling symlink → exit 3, valid links → exit 0) and written up in the commit, while deleting the fix line still left all 101 tests green. Adversarial review caught it; self-review did not.

`rg -l "negative test|mutation test|break the subject" skills/*/SKILL.md` returns nothing. The nearest existing skills — `test-cli-application`, `write-testthat-tests`, `review-codebase`, `review-pull-request` — all describe writing or running checks, none describes proving a check can fail.

## Proposal

A skill, working name `verify-gate-can-fail`, covering the procedure:

1. **Precondition** — the subject file must be clean, since restore is destructive
2. **Baseline** — the suite must be green first; a mutation against a red suite proves nothing
3. **Mutate** — remove or invert the line the check supposedly covers
4. **Verify the mutation landed** — the step everything else depends on. A substitution that matched nothing leaves the file untouched and makes the whole exercise pass vacuously. (`sed`/`perl -0pi` silently no-op on this repo's NTFS mount; bare `grep` resolves to ugrep locally but GNU grep in CI.)
5. **Re-run and record the real exit code** — never through a pipe, or `$?` is the last stage's status
6. **Restore, and verify the restore**
7. **Interpret** — a surviving mutant means uncovered, however convincingly the behaviour was demonstrated by hand. An unexpectedly *large* kill count means collateral coupling and is also worth understanding.

Key distinctions the skill should draw, both learned the hard way:

- **A manual break-and-check proves the feature, not the coverage.** Hand-running the CLI and seeing the right exit code is a demo; "removing this line fails these N tests" is coverage. The manual evidence feels like the negative test was done, which is exactly what makes it dangerous.
- **Scope the assertion.** An exit code reduces over every adapter, so asserting on one couples the suite to global state (`~/.openclaw`, `~/.hermes` are always detected). Sandbox `HOME`; put fixtures in `mkdtempSync`, not under the repo root.

## Tooling already exists

[...]

Issue #459, opened 2026-07-29
ci: nothing consumes the new audit exit code — wire the gate into ci-node-cli.yml


pjt222:
## Gap

#439 gave `audit` a real exit code (PR #458): `0` clean, `2` adapter crashed, `3` findings, with `1` reserved for usage and loader errors. Nothing in CI consumes it yet.

The exit code is currently exercised only by the test suite's own fixtures (`audit exit codes end to end (#439)` in `cli/test/cli.test.js`), which run the CLI against synthetic temp directories. **No CI job runs `audit` against the actual repository checkout.** So a real broken symlink in `.claude/skills/` or `.agents/` — the exact condition #365 was — would still reach `main` without any check objecting.

This was deliberately excluded from PR #458 because a step that can turn `main` red is a separate decision, not because it is unimportant.

## Where it belongs

`.github/workflows/ci-node-cli.yml`. It is the only workflow that can run the CLI at all:

- it does `npm ci` (`:39`), so the CLI's dependencies resolve
- it triggers on `cli/**` (`:6`, `:17`)

**Not** `scripts/validate-integrity.sh`. B11 lived there and was deleted in #443 for exactly two structural reasons, both still true: that workflow performs no `npm ci`, and `cli/**` is absent from its `paths:`. The comment at `scripts/validate-integrity.sh:528-533` says so explicitly and asks that no stdout grep be re-added there. A genuine exit code is the principled replacement B11 never had.

## Proposal

Add a step to `ci-node-cli.yml`:

```yaml
      - name: Audit installed content
        run: node cli/index.js audit
```

A bare invocation is enough — the step fails on any non-zero status, and the full report is printed before the exit code is set, so the log shows *why*.

## Decisions to settle first

1. **Does a clean CI checkout actually audit clean?** Verify before enabling, or the step lands red on day one. A fresh clone has no `.agents/`, and `.claude/skills/` is a committed symlink farm whose targets must resolve on a Linux runner.
2. **Should a crash (`2`) and a finding (`3`) be treated the same?** A bare `run:` treats any non-zero as failure. If a crash should page louder, the step needs to branch on `$?`.
3. **Warn-only first?** `continue-on-error: true` for a period would show whether the step is stable before it can block merges. Given the history here — a gate that could not fire is worse than no gate — a warn-only phase should be time-boxed and then removed, not left indefinitely.

## Acceptance criteria

- [ ] `ci-node-cli.yml` runs `audit` against the repo checkout
[...]

Issue #453, opened 2026-07-29
test(cli): campfire/tend tests mutate the real repo root and never pre-clean, so ambient state decides the result


pjt222:
## Summary

The campfire/gather/tend tests operate on the **real repository root**, not a fixture. Their outcome depends on ambient state that the suite never clears, so the suite can fail spuriously on a developer machine and can leave poisoned state behind when it does.

Same family as the lying-gate work in #363/#365/#443: a check whose result is decided partly by ambient state rather than solely by the subject.

## Mechanism

`cli/test/cli.test.js:33-37`:

```js
const CLI = 'node cli/index.js';
const ROOT = process.cwd();

function run(args) {
  return execSync(`${CLI} ${args}`, { cwd: ROOT, encoding: 'utf8', timeout: 10000 });
}
```

`ROOT` is the repo root, so `run('gather tending')` writes `<repo>/.agent-almanac/state.json` for real. It is gitignored (`.gitignore:25`), so the worktree stays clean — which is exactly why this has gone unnoticed.

## Three distinct defects

**1. No `before()` anywhere — pre-existing state is never cleared.**

Four blocks touch `.agent-almanac`: `campfire` (`:421`), `gather` (`:470`), `scatter` (`:514`), `tend` (`:525`). All four have zero `before()` hooks. So a developer with a real campfire lit fails `tend with no fires shows message` (`:526-529`), which asserts `/No fires to tend/`. The suite is red for reasons unrelated to the change under test.

**2. `tend` cleans up only on success.**

`campfire` and `gather` clean in `after()` (`:424-426`, `:473-475`), which runs regardless of outcome. `tend` instead cleans inline at the *end of each test body* — `:546` and `:564`:

```js
    // Clean up
    try { rmSync(resolve(ROOT, '.agent-almanac'), { recursive: true }); } catch {}
```

An assertion throw at `:543` or `:561` skips the `rmSync` entirely, leaving `.agent-almanac/` behind with `lastWarmed` pinned to the backdated sentinel `2020-01-01T00:00:00.000Z` written at `:539`. The next run then starts dirty — turning one failure into a persistent one.

**3. Concurrent CLI use corrupts the run.**

Because the state dir is the real one, anything else invoking the CLI in that directory during a test run interferes. Observed in practice: a run whose `tend --dry-run` died with

```
Error: ENOENT: no such file or directory, open '/mnt/d/dev/p/agent-almanac/.agent-almanac/state.json'
```

[...]

PR #461, opened 2026-07-29
feat(scripts): add mutation-check; fix --version drift (#456)

### LINKED ISSUE #456: cli: --version reports 0.1.0 against a 1.3.0 package, and cli/README advertises Node 18+ against engines >=22.12.0

pjt222:
Two commits from a maintenance pass: extract the session's repeated procedure into a tool, and fix the version/doc drift it exposed.

Closes #456.

## 1. `scripts/mutation-check.js` (`6ff9a960`)

The repo's most load-bearing discipline — *a green gate is evidence about the gate, not the subject* — has been applied by hand every time, and it is easy to get wrong while believing you are doing it.

In #458 the audit exit code was verified end to end manually (dangling symlink → exit 3, valid links → exit 0) and written up in the commit, yet deleting `process.exitCode = auditExitCode(results)` still left all 101 tests passing. The manual evidence *felt* like the negative test had been done. Adversarial review caught it; self-review did not — in a PR whose subject was gates that cannot fail.

The tool runs the envelope:

1. refuse to run if the target has uncommitted changes (restore is `git checkout --`, which would discard them)
2. require a green baseline — a mutation against a red suite proves nothing
3. apply the mutation
4. **verify via `git diff` that it actually landed**
5. re-run, recording the real exit code, never through a pipe
6. restore, and verify the restore happened

Step 4 is what makes the rest trustworthy. A substitution that matched nothing leaves the file untouched and turns every later result into a lie — the failure the #442 session hit, where a no-op made a negative test pass against an unmodified file.

Node rather than shell, deliberately: in-place `sed`/`perl -0pi` silently no-op on the NTFS mount, and bare `grep` resolves to ugrep locally but GNU grep in CI. Both are failure modes the tool exists to catch, so it depends on neither.

**Dogfooded on the case that motivated it:**

```
$ npm run mutation-check -- --file cli/index.js \
    --delete-matching 'process.exitCode = auditExitCode' \
    --test 'npm run test:cli' --expect-killed-by 2
[2/4] applying mutation ...
      landed: cli/index.js | 1 -
[3/4] running tests against the mutant (expect red) ...
      exit 1, 2 failing
MUTANT KILLED by 2 failing test(s) — the check can fail.
```

Guards verified too: an untracked file is refused, and a non-matching pattern aborts with *"the mutation did not change the file"* rather than reporting a result. Working tree left clean in both cases.

## 2. Version and Node-floor drift (`99a86089`)

`--version` was hardcoded and had drifted **three** ways: the binary printed `0.1.0`, `cli/README.md` advertised `1.1.0`, and the published package is `1.3.0`. Every user running `--version` — including anyone answering "what version are you on?" in a bug report — got a wrong answer, on every release since `0.1.0`.

Now reads `pkg.version` from the manifest, so it cannot drift again.

The test is why this survived. It asserted `/\d+\.\d+\.\d+/`, which `0.1.0` satisfies — a check that ran on every release and was structurally unable to detect the defect it appeared to cover. It now compares against `package.json`, re-reading the manifest rather than importing the same module the CLI does, so if the version wiring breaks the test still knows the truth.

**Verified with the tool from commit 1** — reverting `pkg.version` to `'0.1.0'`:

```
MUTANT KILLED by 1 failing test(s) — the check can fail.
```

The old assertion would have stayed green under that same mutation.

Also corrects `cli/README.md`'s "Requires Node.js 18+" to 22.12, matching `engines`. `engines` is advisory — npm warns rather than blocks — so the README was steering users into an unsupported runtime rather than being caught at install time. The hardcoded example output under `--version` is removed rather than updated, since restating the version in prose is exactly what drifted.

## Verification

103/103, unchanged in count: the version test was strengthened in place, not added.

## Related

Skill proposal for the procedure this automates: #460. Also filed this pass: #459 (nothing consumes the new audit exit code yet), and #453 broadened with observed evidence.

Not filed, because the premise did not survive checking: the standing note that `validate-integrity.yml` lacking a `cli/**` trigger is a gap. After B11's removal, `scripts/validate-integrity.sh` no longer reads `cli/` at all, and the comment at `:528-533` states that audit coverage lives in `ci-node-cli.yml` by design. The note is stale, not a bug.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -72,6 +72,27 @@ Guides, skills, agents, and teams are cross-referenced. The parent project `CLAU
   done
   ```
 
+## Proving a Gate Can Fail
+
+A green check is evidence about the *check*, not about the subject. Before trusting a gate — and always before claiming to have tightened one — break the thing it guards and confirm the gate goes red.
+
+`scripts/mutation-check.js` automates the envelope:
+
+```bash
+npm run mutation-check -- \
+  --file cli/index.js \
+  --delete-matching 'process.exitCode = auditExitCode' \
+  --test 'npm run test:cli' \
+  --expect-killed-by 2
+```
+
+It refuses to run on a file with uncommitted changes (restore is `git checkout --`), requires a green baseline, **verifies via `git diff` that the mutation actually landed**, records the real exit code, then restores and verifies the restore. Exit 0 means the mutant was killed; exit 1 means it survived and the line is uncovered.
+
+Two traps this exists to catch, both of which have shipped here:
+
+- **A mutation that silently matched nothing** makes the whole exercise pass vacuously while looking correct. In-place `sed`/`perl -0pi` no-op on the NTFS mount, and bare `grep` resolves to ugrep locally but GNU grep in CI — which is why the tool is Node and checks `git diff` rather than trusting the edit.
+- **A manual break-and-check proves the feature, not the coverage.** Running the CLI by hand and seeing the right behaviour is a demo; "removing this line fails these N tests" is coverage. In #458 the exit code was verified end to end by hand and written up in the commit, while deleting the fix line still left all 101 tests green.
+
 ## Adding a New Skill
 
 1. Create `skills/<skill-name>/SKILL.md` following the format of existing skills
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -86,11 +86,14 @@ npm run mutation-check -- \
   --expect-killed-by 2
 ```
 
-It refuses to run on a file with uncommitted changes (restore is `git checkout --`), requires a green baseline, **verifies via `git diff` that the mutation actually landed**, records the real exit code, then restores and verifies the restore. Exit 0 means the mutant was killed; exit 1 means it survived and the line is uncovered.
+Exit 0 means the mutant was killed and the check works. Exit 1 means it survived — the line is uncovered — or the run could not produce an honest verdict.
 
-Two traps this exists to catch, both of which have shipped here:
+The tool refuses to guess. It requires a green baseline, backs the file up on disk before touching it, restores from an in-memory buffer (never `git checkout --`, which restores from the *index*), and reports `INCONCLUSIVE` rather than a kill whenever the test command could not be interpreted. It declines to run at all on a symlink, on an `--assume-unchanged`/`--skip-worktree` file, on a file with uncommitted changes, on a mutation matching more than one site without `--allow-multiple`, or while a stale `.mutation-check.bak` is present.
 
-- **A mutation that silently matched nothing** makes the whole exercise pass vacuously while looking correct. In-place `sed`/`perl -0pi` no-op on the NTFS mount, and bare `grep` resolves to ugrep locally but GNU grep in CI — which is why the tool is Node and checks `git diff` rather than trusting the edit.
+Three traps it exists to catch, all of which have shipped here:
+
+- **A mutation that silently matched nothing** makes the exercise pass vacuously while looking correct. In-place `sed`/`perl -0pi` no-op on the NTFS mount, and bare `grep` resolves to ugrep locally but GNU grep in CI — which is why the tool is Node and compares content in memory rather than trusting the edit.
+- **A mutation that merely breaks parsing is not coverage.** Deleting a line carrying a brace makes the file fail to load, and node:test reports that as one failing test — indistinguishable from a real kill. Mutants are syntax-checked first and reported `INVALID` if they do not parse. Note `node --check` parses a `.js` file as CommonJS, so the check is done through a temp file whose extension matches the package's module type; without that this guard is dead for every `.js` file in an ESM package.
 - **A manual break-and-check proves the feature, not the coverage.** Running the CLI by hand and seeing the right behaviour is a demo; "removing this line fails these N tests" is coverage. In #458 the exit code was verified end to end by hand and written up in the commit, while deleting the fix line still left all 101 tests green.
 
 ## Adding a New Skill
--- a/cli/README.md
+++ b/cli/README.md
@@ -66,10 +66,14 @@ npm install -g agent-almanac
 This registers two commands: `agent-almanac` and `almanac` (shorthand). Verify with:
 
 ```bash
-agent-almanac --version   # 1.1.0
+agent-almanac --version
 ```
 
-Requires Node.js 18+.
+The version is read from the package manifest, so it always matches the
+installed release. (No example output here on purpose — the previous hardcoded
+number drifted from both the manifest and the binary. See #456.)
+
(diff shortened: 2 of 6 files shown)


pjt222:
## Adversarial review round (`f82a09c5`)

The tool written to prove that gates can fail was itself an ungated gate. Copilot is still quota-blocked, so this went through the substitute path — two attack lenses (data loss, false verdict), each finding handed to a separate agent instructed to reproduce it or mark it refuted.

**13 raised, 11 confirmed, 4 blocking.** Two destroyed uncommitted work. Merging on green CI would have shipped them.

### Data loss

| Defect | Fix |
| --- | --- |
| `git update-index --assume-unchanged` makes git report a file clean however far the worktree has diverged. The precondition passed on a file full of uncommitted work, `git diff --quiet` then said "nothing matched", and `git checkout --` overwrote it from HEAD. Work gone, and the only message on screen was a wrong diagnosis. | Refuse such files, **and remove `git checkout --` entirely** — restore is now purely from the in-memory buffer, so git is never asked to write to the worktree |
| `writeFileSync` followed a symlinked `--file` and wrote through to a target git has no copy of, momentarily truncating it. A poller observed three states including a zero-length window; SIGKILL there left the target permanently damaged with `git status` clean. | Refuse symlinks |
| `git checkout --` restores from the **index**, not HEAD, so a `--test` command that stages files left the mutation staged while the tool printed success | Same removal as above |

### False verdicts

| Defect | Fix |
| --- | --- |
| A mutation that merely broke JS parsing was reported as `MUTANT KILLED ... the check can fail`. Deleting a line carrying a brace makes the file fail to load, which node:test counts as one failing test — the strongest possible claim, printed for a line with **zero** coverage. The #458 failure mode dressed up as a pass. | Syntax-check the mutant; report `INVALID MUTANT` |
| `spawnSync`'s 1 MiB default maxBuffer SIGTERMs the child and returns `status: null`; `?? 1` turned a genuinely **green** run into a kill | 64 MiB, and `error`/`signal`/127 inspected → `INCONCLUSIVE` |
| Nonexistent `--test` command counted as a kill | baseline always runs; 127 detected |
| `--expect-killed-by` silently ignored when the count could not be parsed | errors instead |
| `--delete-matching` deleted **every** matching line, so a collateral site could earn a kill credited to the target line | refuse >1 site without `--allow-multiple` |
| `--replace` mutated only the first occurrence → false `SURVIVED` | replaces all, with the same site guard |
| Empty `--delete-matching` blanked the file and "killed" it | rejected |

### The guard that needed its own negative test

The parse guard failed its first negative test. `node --check` parses a `.js` file as **CommonJS**, so a mangled ESM file exits 0 as `.js` and 1 as `.mjs` — measured. The guard was inert for every file in `cli/`, which is the entire subject area. It now probes through a temp file whose extension matches the package's module type, and is verified in both directions: a brace deletion reports `INVALID MUTANT`, a value change reports `it parses` and proceeds to a real verdict.

Worth stating plainly, since it is the same lesson twice in one PR: writing the guard is not the same as knowing it can fire.

### Verification

Every fix checked by re-running the reviewers' own reproductions, not by reasoning:

```
1. assume-unchanged  -> ERROR: carries the git index flag 'h' ...   work SURVIVED
2. symlink           -> ERROR: is a symlink ...                     target SURVIVED
3. syntax break      -> INVALID MUTANT — the mutated file does not parse
4. missing command   -> ERROR: Baseline inconclusive — not found (exit 127)
5. empty needle      -> ERROR: --delete-matching needs a non-empty string
6. multi-site        -> ERROR: affects 3 sites, not 1
7. genuine kill      -> MUTANT KILLED
8. genuine survival  -> MUTANT SURVIVED
9. scratch tree      -> clean, no .bak left behind
```

Original dogfood unchanged: `MUTANT KILLED by 2 failing test(s)` on the #458 case, no backup residue. 103/103.

### Recovery, and one residual

The original is written to `<file>.mutation-check.bak` **before** the file is touched, and a later run refuses to start while a stale backup exists, printing the recovery command. That covers `SIGKILL`, which no handler can trap.

Residual, filed as **#462**: `Ctrl-C` is ignored during the test run because `spawnSync` blocks the event loop. Mitigated by the backup — failure mode is "confusing, recoverable" rather than silent loss — so it is a follow-up, not a blocker.

Two findings were refuted: a claimed missing recovery breadcrumb (the backup covers it) and a restore-warning false positive under `--skip-worktree` (now refused outright).

PR #466, opened 2026-07-30
fix(scripts): run the mutant test asynchronously so Ctrl-C works (#462)

### LINKED ISSUE #462: scripts: mutation-check ignores Ctrl-C during the test run because spawnSync blocks the event loop
## Gap

`scripts/mutation-check.js` installs `SIGINT`/`SIGTERM` handlers to restore the mutated file, but it runs the test command with `spawnSync`. A synchronous child blocks the event loop, so **no signal handler can fire for the entire duration of the test run** — which is the longest phase, minutes on this mount.

Consequences:

1. **Ctrl-C appears to do nothing.** The user presses it, sees no response, and reasonably escalates to `kill -9`.
2. A `SIGKILL` is untrappable, so the escalation strands the mutant on disk.

## Why it is not currently dangerous

Mitigated in the hardened version: the original content is written to `<file>.mutation-check.bak` **before** the file is touched, and a subsequent run refuses to start while a stale backup exists, printing the exact recovery command:

```
ERROR: A backup from a previous run is still present:
  cli/index.mutation-check.bak
That run did not finish cleanly, so the file may still hold a mutation.
Recover with:  mv "<backup>" "<file>"
```

So the failure mode is "confusing, recoverable" rather than "silent data loss". That is why this is a follow-up and not a blocker.

Also worth noting: `SIGINT` from a terminal goes to the whole process group, so the child dies too and the parent's handler *does* run once `spawnSync` returns. The dead zone is narrower than it looks — but it is real for a signal sent directly to the parent PID.

## Fix

Switch the test invocation to async `spawn` with an awaited promise, so the event loop stays live and handlers fire promptly. `runCommand()` is the only place that would change; everything downstream already works from a `{ status, signal, error, output }` shape.

Care needed on two points, both of which the sync version gets right today and a rewrite could regress:

- **Do not reintroduce the maxBuffer trap.** `spawnSync`'s 1 MiB default returned `status: null` on overflow, which a `?? 1` coercion turned into a fake kill. Streaming `spawn` avoids the cap, but the accumulated output still needs a bound so a runaway suite cannot exhaust memory.
- **`error` and `signal` must stay distinguishable from a real non-zero exit**, or the tool resumes reporting "I killed the suite myself" as "the suite went red".

## Acceptance criteria

- [ ] Ctrl-C during the test phase restores the file and exits promptly
- [ ] **Negative test:** send `SIGINT` to the parent mid-run and confirm the file is restored and no `.bak` remains — verify, do not reason
- [ ] `SIGKILL` still leaves a recoverable `.bak` and the stale-backup guard still fires
- [ ] maxBuffer-overflow and command-not-found cases still report INCONCLUSIVE, not a kill

Relates #460. Found by adversarial review of the tool in PR #461, which reproduced the blocked-handler behaviour directly.

pjt222:
Closes #462.

`mutation-check` installed `SIGINT`/`SIGTERM` handlers but ran the test command with `spawnSync`. A synchronous child blocks the event loop, so no handler could fire for the whole test run — the longest phase, minutes on this mount. Ctrl-C appeared to do nothing, and the reasonable next move is `kill -9`, which is untrappable and strands the mutant on disk.

`runCommand()` now uses async `spawn` with an awaited promise.

## The defect the review caught before it shipped

The plan I started from converted `runCommand` to async and **did not touch either call site**. `runCommand` would have returned a Promise, `.status` would have been `undefined`, `inconclusiveReason()` would have seen no error and no signal, and `status !== 0` would have been true for *every* run — reporting the mutant killed regardless of what the tests did.

A false green in the one tool whose entire job is detecting false greens. Both `await`s are in.

## Preserved deliberately

| Property | Why it matters |
| --- | --- |
| Output is bounded (`MAX_OUTPUT_CHARS`) | `spawnSync`'s `maxBuffer` capped output by SIGTERM-ing the child and returning `status: null`, which an earlier `?? 1` turned into a fake kill. Streaming `spawn` has **no** cap, so the bound is applied here and an overflow is reported as `error` (`ENOBUFS`) — never as a verdict |
| `error` / `signal` stay separate from a real non-zero exit | So `inconclusiveReason()` can still tell "the suite went red" from "I killed the suite myself" |

## Comments corrected

Two became false with this change and are fixed rather than left to rot:

- the note claiming the handlers "fire only between phases"
- `inconclusiveReason()`'s "could not be run", which is self-contradictory for an overflow — the command ran, it just did not finish

## Verification

In a scratch git repo, never the real one:

```
normal operation      MUTANT KILLED, restored, no .bak
command not found     "Baseline inconclusive — not found (exit 127)"
SIGINT mid-test-run   exit 130 in 6 ms, file restored, no stray .bak
```

The 6 ms is the whole point: that window used to be the entire test run.

I did not measure the old binary's SIGINT latency directly — the mechanism (`spawnSync` blocking the loop) and the new measurement are the evidence, and I'd rather say that than imply a before/after benchmark I didn't run.

## Out of scope, stated

The on-disk backup stays: `SIGKILL` is still untrappable, and a stale `.mutation-check.bak` still blocks the next run with a recovery command.

Killing an **orphaned test child** — reachable only when `SIGINT` goes to the parent PID alone rather than the process group, as a terminal does — needs a module-level handle to the in-flight child, and doing it properly needs `detached: true` plus `process.kill(-pid)`, which changes signal delivery for the ordinary terminal case too. Left out; say if you want it.

### DIFF
--- a/scripts/mutation-check.js
+++ b/scripts/mutation-check.js
@@ -37,7 +37,7 @@
  * silently no-op on this repo's NTFS mount, and bare `grep` resolves to ugrep
  * locally but GNU grep in CI — the two failure modes this tool exists to catch.
  */
-import { spawnSync } from 'node:child_process';
+import { spawn, spawnSync } from 'node:child_process';
 import { existsSync, lstatSync, readFileSync, unlinkSync, writeFileSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { dirname, extname, relative, resolve } from 'node:path';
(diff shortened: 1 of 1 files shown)


pjt222:
Dogfooded against the real repo and the real suite, not only the 2-line scratch fixture:

```
$ npm run mutation-check -- --file cli/index.js \
    --delete-matching 'process.exitCode = auditExitCode' \
    --test 'npm run test:cli' --expect-killed-by 2
[3/5] checking the mutant still parses ... it parses.
[4/5] running tests against the mutant (expect red) ...
MUTANT KILLED by 2 failing test(s) — the check can fail.
```

Same verdict and same kill count as the spawnSync version, so the async conversion is behaviour-preserving on the path that matters. Working tree clean afterwards, no stray `.mutation-check.bak`.

Issue #272, opened 2026-06-04
Normalization pass + cli-rs renderer recheck for content styleguide (follow-up to #258)


pjt222:
Follow-up to #258, which established the canonical content styleguide (`guides/content-styleguide.md`) and a forward-enforcing CI check. This issue tracks the deferred **retroactive** work that #258 explicitly scoped as "(2) a follow-up normalization pass".

## Why deferred
The styleguide PR (#258 branch `feat/258-content-styleguide`) deliberately did **not** rewrite the existing corpus. A repo-wide normalization is an unreviewable diff: most English `*.md` blobs are CRLF and `.gitattributes` mandates `eol=lf`, so every touched file renders as a full-file CRLF→LF change on top of the content edit. Mixing that churn with semantic rewrites would bury the real changes.

## Scope (measured by `scripts/check-content-style.js --all`)
- **Decorative table separators**: ~1,589 across ~931 files. Mechanical/scriptable — rewrite `|----...|` separator rows to compact `|---|`.
- **Untagged code fences**: ~5,973 across ~1,974 files. **Not** auto-inferable — each block needs a per-block language guess (`bash`/`r`/`yaml`/`json`/`python`/`text`/…). Good fan-out/agent job, not one-shot `sed`.

## Tasks
- [ ] Dedicated normalization PR: `git add --renormalize '*.md'` (CRLF→LF) **plus** the table-separator + fence-tag rewrites, so the line-ending churn and the content normalization land together and review cleanly.
- [ ] Flip the untagged-fence CI rule from **warn** to **block** in `validate-content-style.yml` once the corpus is clean (uses the existing full-file fence-state parser).
- [ ] Refresh the i18n translations staled by #258's English cross-ref edits (notably the 10 `create-skill` locale copies — currently warn-only stale).
- [ ] Re-check the `cli-rs` markdown renderer (`cli-rs/src/content/markdown.rs`, branch `rust-cli`) against the normalized corpus — confirm no remaining table/fence mangling (#255).

## References
- Styleguide: `guides/content-styleguide.md`
- Checker: `scripts/check-content-style.js` (`--all` for the repo-wide count, `--added <base>` for PR scope)
- Parent: #258 · Rust port: #255

pjt222:
## AC#4 complete: cli-rs renderer recheck against the normalized corpus

Done on branch `rust-cli` (merge `e55998fe` brought the normalized corpus in; results commit `d39a64f5`).

### Method

[...]

PR #516, opened 2026-08-10
fix(content): repair five runaway fences in write-incident-runbook EXAMPLES (#511)

### LINKED ISSUE #511: content: four runaway fences in write-incident-runbook/references/EXAMPLES.md
Split out of #510 (the #476 repair) as an isolated case with a more tangled shape than the eight files fixed there.

## Defect

`skills/write-incident-runbook/references/EXAMPLES.md` has four `text` fences that have swallowed the document's own structure:

```
L204 `text` 15 lines   swallows "## Step 2: Complete Diagnostic Procedures"
L333 `text` 28 lines
L537 `text` 15 lines
L571 `text` 64 lines
```

122 lines render as preformatted text.

## Why it is harder than the eight fixed in #510

Those were a single `markdown` template with one nested fence. Here the L9 template nests **two different tagged fences** (`promql` at L45, `logql` at L52), each closing with a bare ``` that shuts the outer template early:

```
  9: ```markdown
 45: ```promql
 49: ```          <- closes the TEMPLATE, not the promql block
 52: ```logql
 55: ```
...
204: ```text      <- runaway
```

Four-backticking the outer template is still the fix, but the intended closer has to be identified per block rather than pattern-matched, because the early-close point and the runaway opener are not adjacent.

## Not mirrored

`references/EXAMPLES.md` files are not translated — `i18n/<locale>/skills/write-incident-runbook/` contains only `SKILL.md`. So this is 1 file, not 1 + 10, and fixing it has no re-mirror cost. That is why it was safe to defer rather than block #510.

## Detection

The detector used on #510 keys on three signals: `**Expected:**` / `### Step N` inside a non-`markdown` fence; an H2 inside a `text`/untagged fence; or a fence that runs to EOF opening right after a closing delimiter. This file trips the second.

Note the H2 signal must be restricted to `text` and untagged fences — `##` is a comment marker in bash/R/python, and `markdown` fences legitimately show templates with headings. Dropping it outright (rather than restricting it) is what let a runaway ship in #510's first pass.

## Related

#476 (runaway fences, fixed for the other 8 files in #510) · #472 (the frozen-fence rule) · #501 (parked-set register)

pjt222:
Closes #511.

`skills/write-incident-runbook/references/EXAMPLES.md` had four `text` fences that had swallowed the document's own structure — **122 lines rendering as preformatted text**, burying `## Step 2: Complete Diagnostic Procedures` and `## Step 4: Complete Escalation Guidelines`.

## Root cause, from git rather than inference

Every one of these delimiters was a bare ``` in the file as its author wrote it. `33b561c96` ("style(content): tag untagged code fences", #272) rewrote **17** bare delimiters to ```` ```text ```` — `main` carries 18 such lines, one of which predates that pass. Wherever its parser had already been thrown by a nested tagged fence, it tagged a **closer** as an opener.

Line 204 is the clearest case — it closes the runbook template opened at line 9, and sits directly under a Revision History table with nothing after it to open a block for:

```
   9: ```markdown        <- template opens
  45: ```promql          <- nested
  49: ```                <- closes the TEMPLATE, not the promql block
 204: ```   ->  ```text  <- tagger now believes it is at document level
```

## The repair

Rather than patch delimiters one at a time across 920 lines, this rebuilds from `33b561c96^` — the author's own delimiters — and raises the **six** ```` ```markdown ```` templates that *nest* a fence to four backticks, opener and closer. Inside a four-backtick fence the nested ``` are literal content and can no longer close the template early.

Every bare delimiter in this file then turns out to be interior to a template, so #272's "tag untagged fences" rule has nothing to do here at all. Of `main`'s 18 `text` lines, **14 revert to bare** and **4 become four-backtick closers** (204, 333, 537, 571); none remain.

**Nothing landed on this file between the tagging pass and this PR** (`git log` shows `33b561c96` as the newest commit touching it), so rebuilding from its parent reverts nothing.

## Two runaways the issue did not list

#511 names four (204, 333, 537, 571), found by the #510 detector's "H2 inside a `text` fence" signal.

- **Line 643** is the same defect — a tagged closer of the escalation-message block opened at 636 — missed because its fence contains only an H3.
- **The template at line 688** nests nine blocks and runs to **867**, not 694. A stack-based reading closes it at the first bare ```; only the content settles it, since line 689 is the template's own `## Communication Templates` heading. Leaving it unraised inverted every delimiter in that region — line 708 follows "Next update in 15 minutes." and line 866 follows "Best regards,", and both had been tagged as if they opened something.

The tell for that second one was an **empty `text` fence at line 866** in the first pass — the signature of an off-by-one, where 866 closes the last nested block and 867 closes the template.

This is why the issue asked for the closer to be "identified per block rather than pattern-matched": every automatic reading of these delimiters, including the one I wrote first, gets this template wrong.

## Measured

| | before | after |
|---|---:|---:|
| fences | 57 | 9 |
| runaway fences (H2 buried in a text/untagged fence) | 4 | **0** |
| untagged fences | 0 | 0 |
| empty fences | 1 | **0** |
| fenced lines | 391 | 860 |
| buried `## Step N` headings | 2 | **0** |

Fence count falls because nested delimiters are now literal content — the file is 7 `markdown` templates plus one `yaml` and one `json` block. Fenced lines rise because the templates finally hold their full bodies. The 122 preformatted lines the issue counted are 15 + 28 + 15 + 64, exactly the four runaway bodies.

**25 lines changed, every one a delimiter.** Content is byte-identical to `main` once delimiter lines are excluded — verified by diffing both revisions with delimiters filtered out.

Not mirrored — `i18n/<locale>/skills/write-incident-runbook/` holds only `SKILL.md` — so this is 1 file with no re-mirror cost.

`validate:integrity` passes; `validate:line-endings` clean; content-style clean on added lines.

### DIFF
--- a/skills/write-incident-runbook/references/EXAMPLES.md
+++ b/skills/write-incident-runbook/references/EXAMPLES.md
@@ -685,7 +685,7 @@ Escalate **proactively** if:
 
 ### Internal Communication Templates (All Variants)
 
-```markdown
+````markdown
 ## Communication Templates
 
 ### Initial Incident Declaration
@@ -705,7 +705,7 @@ Escalate **proactively** if:
 Quick Summary: [1-2 sentences on what's wrong and initial assessment]
 
 Next update in 15 minutes.
-```text
+```
 
 ### Progress Update (Every 15-30 minutes)
 
@@ -725,7 +725,7 @@ Next update in 15 minutes.
 **Impact Update**: [Any change in scope/severity]
 
 Next update in 15 minutes or upon significant change.
-```text
+```
 
 ### Mitigation Announcement
 
@@ -743,7 +743,7 @@ Next update in 15 minutes or upon significant change.
 
 Continuing to monitor for 30 minutes before declaring resolved.
 Next update in 15 minutes.
-```text
+```
 
 ### Incident Resolution
 
@@ -765,7 +765,7 @@ Next update in 15 minutes.
 **Dashboard**: [Link to incident metrics]
 
 Thank you to everyone who helped resolve this incident.
-```text
+```
 
 ### False Alarm / Resolved Without Action
 
@@ -779,7 +779,7 @@ Thank you to everyone who helped resolve this incident.
 
 No customer impact.
 No follow-up needed.
-```text
+```
 
 ### External Status Page Updates
 
@@ -796,7 +796,7 @@ Our engineering team has been notified and is actively investigating.
 **Next Update**: [HH:MM UTC] (15 minutes)
 
 We apologize for any inconvenience.
-```text
+```
 
 **Progress Update**:
 ```
@@ -809,7 +809,7 @@ Our team is implementing a fix.
 **Impact**: [Percentage or description of affected users]
 **Estimated Resolution**: [Timeframe if known, or "working on it" if not]
 **Next Update**: [HH:MM UTC]
-```text
+```
 
 **Resolution Post**:
 ```
@@ -826,7 +826,7 @@ All services are operating normally.
 We have implemented measures to prevent this from happening again and will continue monitoring closely.
 
 We apologize for any inconvenience this may have caused.
-```text
+```
 
 ### Customer Support Email Template
 
@@ -863,8 +863,8 @@ If you have any questions, please contact our support team at [email/phone].
 Best regards,
 [Name]
 [Title]
-```text
 ```
+````
 
 ## Step 6: Alert Integration Examples
 
--- a/skills/write-incident-runbook/references/EXAMPLES.md
+++ b/skills/write-incident-runbook/references/EXAMPLES.md
@@ -6,7 +6,7 @@ Complete configuration files, templates, and multi-variant examples extracted fr
 
 ### Advanced SRE Runbook Template (Complete)
 
-```markdown
+````markdown
 # [Service Name] - [Incident Type] Runbook
 
 ## Metadata
@@ -116,7 +116,7 @@ Confirm incident is resolved and system is healthy.
 **Slack Channel**: #incident-response
 
 **Initial Message Template**:
-```text
+```
 🚨 INCIDENT: [Title]
 Severity: [Critical/High/Medium]
 Impact: [Description]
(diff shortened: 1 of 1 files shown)


pjt222:
**Adversarial review: three lenses, all `correct`, no defect found in the file.**

Copilot did not pick this PR up and the advocatus-diaboli agent idled four times without producing a report, so this ran as a three-lens workflow review — boundaries, content-preservation, and a render-as-a-reader pass. Recording it here so the provenance is not "unreviewed".

Independent corroboration worth keeping:

- A CommonMark scan resolves exactly the nine claimed fences at exactly the claimed lines. Every one of the 9 openers is preceded by blank + an H3, and every one of the 9 closers is followed by blank + an H2/H3 or EOF — nine-for-nine in both directions.
- `diff 33b561c96^ HEAD` is **exactly 12 lines**, all delimiter promotions. With delimiter lines stripped, the 798 content lines are byte-identical to **both** `origin/main` and `33b561c96^`.
- Interior fence counts per template are all even (16/28/0/34/6/2/18/0/0), so no nested snippet is left half-open.
- The repair also **revives six dead anchors**: `SKILL.md` links in at `#step-N-…` from lines 46/88/138/191/229/280, and two of those targets were trapped inside runaway fences on `main`.

Seven findings, all low severity, and all of them were arithmetic in this PR's own prose rather than defects in the file. Three verified and corrected above:

| claim | was | actual |
|---|---|---|
| tags introduced by #272 | 17 | 17 added by that commit; `main` carries 18, one predating it |
| disposition of those tags | "all 13 revert to bare" | 14 → bare, 4 → four-backtick closers |
| empty fences, before | 0 | 1 (at 866 on `main`) |

Two further findings are **pre-existing content asymmetries**, byte-identical to the author's original and out of scope for a delimiter repair — noting them so they are not lost:

- **Line 681** — the Cloudflare vendor entry ends after `**Account**`, while AWS and MongoDB both carry a third `**Support Portal**` field. It sits on the last line before a fence boundary, which is exactly where a rebuild would drop a line, so it deserved the scrutiny; the 12-line diff settles that it is original.
- **Line 686** — `### Internal Communication Templates (All Variants)` introduces a template whose second half is external-facing (`### External Status Page Updates` at 784, `### Customer Support Email Template` at 831). `SKILL.md:229` describes the same target as "all internal and external templates", so the subheading is the narrower of the two.

I will file the second one as a follow-up rather than widen this PR.

Issue #501, opened 2026-08-07
i18n: parked-set register — everything #477 defers to human review


pjt222:
Running register of what the #477 mechanical batches deliberately do **not** touch, so it can be reviewed as a whole rather than rediscovered per batch. Updated as batches land.

State on `main` at `111f16d9` (after batches #495 `yaml`+`json`, #496 `bash`, #499 `r`): **468 findings remain**, partitioning exactly:

| category | findings | files | disposition |
|---|---:|---:|---|
| mechanically restorable, not yet batched | 175 | 60 | continue tag-scoped batches |
| ordinal mapping unsound — normalizer skips | 206 | 46 | #478 / #483 |
| `agents`/`teams`/`guides` mirrors | 87 | 11 | #477 AC — normalizer does not cover |
| **total** | **468** | | |

The parked items below are drawn from the 175 and are excluded from every mechanical batch.

---

## 1. Regulated content — needs a named human reviewer

**23 findings across 6 files.** All are mechanically restorable; that is exactly why they must be pulled out. Carve-out keys on `domain: compliance` in the English skill, not on a hand-written list — two of these (`validate-statistical-output`, `write-validation-documentation`) appeared in no list until the predicate was applied.

| file | findings |
|---|---|
| `zh-CN/implement-audit-trail` | r=5 bash=1 |
| `zh-CN/validate-statistical-output` | r=6 |
| `zh-CN/implement-pharma-serialisation` | python=3 sql=1 |
| `zh-CN/setup-gxp-r-project` | r=2 bash=1 |
| `wenyan-ultra/setup-gxp-r-project` | r=2 bash=1 |
| `zh-CN/write-validation-documentation` | r=1 |

**Why file-scoped and not fence-scoped.** These files' fences spread across four tag batches. Letting each batch take its own slice would hand the reviewer a file already partly rewritten — the requirement met for most fences and silently bypassed for the rest, with nothing going red. See the durable form of this in the carve-out rationale on #496.

**The specific line that motivates the gate:** `i18n/zh-CN/skills/implement-audit-trail/SKILL.md` logs `ifelse(match, "通过", "失败")` where English logs `"PASS"`/`"FAIL"` — the recorded value of a 21 CFR Part 11 audit record. Reverting to English is almost certainly correct, but it changes what a regulated record contains, which is not a call a bulk diff should make.

[...]

Issue #478, opened 2026-07-31
i18n: 33 translations were re-authored from metadata by a failed campaign batch — delete and re-scaffold (Class A)


pjt222:
## What these are

**33 translated skills that were re-authored from metadata rather than translated.** They
were written from each skill's `name`, `description` and section template, not from its
body — the source never loaded.

They are *not* abridgements, not staleness, and not independently-authored alternatives.
The original framing of this issue ("coherent documents on the same topic, arguably
better than a faithful translation") does not survive the evidence, and the probe that
established that is in the comment thread below.

## Root cause, from the batch's own commit message

`00f1206c` (2026-03-17):

> All 7 waves initiated in parallel but **agents hit rate limits before completion**.

A wave was a **(locale × domain)** slice. Where the source load failed, every file in the
slice was generated from frontmatter + template; where it succeeded, output is faithful.
**5 of the 8 affected batches are complete domain sweeps:**

```
de / citations          3 of 3      es / citations         3 of 3
ja / citations          3 of 3      zh-CN / citations      3 of 3
es / containerization  10 of 10
de / shiny 5 of 7   |   de / web-dev 3 of 4   |   es / mcp-integration 3 of 5
```

Survivors of the partial batches are clean, not marginal — `de/deploy-shinyproxy`,
`de/design-shiny-ui`, `de/deploy-to-vercel`, `es/troubleshoot-mcp-connection` all sit at
100% code-token survival with exact fence parity.

The forks cluster by **batch**, not by skill: **21 of the 24 affected skills forked in
exactly one locale.** `citations` is a 3-skill domain, which is the only reason four
locales appear to "agree" on it.

Why the failure produced documents rather than nothing: `scripts/translate-content.sh:78`
(`cp "$SOURCE_FILE" "$TARGET_FILE"` — scaffold-then-edit) landed at `77006e7a`, four days
before the campaign, and was not used for these. The later caveman/wenyan campaign did
scaffold first and achieved fence parity on the same English sources.

## The files

**Three are invisible to the gate** (zero gated fences → zero divergences, so
`normalize-i18n-fences.js` never reports them — same mechanism as #480). They are marked
`†` and must be handled explicitly or they will be missed.

**de (11)** — `build-shiny-module`, `deploy-shiny-app`, `format-citations`†,
`manage-bibliography`, `optimize-shiny-performance`, `scaffold-nextjs-app`,
[...]

Issue #477, opened 2026-07-31
i18n: clear the 1,220-fence parity backlog and flip check-i18n-fence-parity to blocking


pjt222:
## Tracking

`scripts/check-i18n-fence-parity.js` ships warn-only because the corpus is not clean:
**1,307 frozen-fence violations across 338 files** at introduction (1,220 in skills, 87 in the agents/teams/guides mirrors). This issue tracks
clearing them and flipping the gate to blocking.

The warn state is the risk. #472's own root cause is a rule stated in four places that
nothing enforced; a warn stream nobody watches is the same failure with extra steps.
This issue is the named exit.

## Split (measured, `node scripts/normalize-i18n-fences.js --dry`)

| | fences | files |
|---|---|---|
| mechanically restorable (ordinal mapping sound) | 1,014 | 281 |
| mirrors — normalizer does not yet cover | 87 | 11 |
| content forks — hand repair, tracked in #478 | 206 | 46 |
| **total** | **1,307** | **338** |

By tag: bash 574, r 320, yaml 198, python 94, dockerfile 37, typescript 27, json 15,
nginx 9, javascript 7, protobuf 7, go 6, css 5, latex 2, gitignore/language/gotmpl/
promql/html/sql 1 each.

## Order

1. **#476 first** — runaway English fences change which lines parse as fences; 18 of
   the 1,220 sit inside runaway regions. Re-measure after it lands.
2. **#475 first** — patch `bulk-scaffold-caveman.sh` before restoring, or the next
   scaffold run re-injects. (Patched in the #472 PR.)
3. Mechanical restore in **tag-scoped batches**, one PR each, most-mechanical first:
   `yaml`+`json` (206, includes all 64 scaffolder injections (#475)), then
   `bash`/`sh`/`console` (510), then `r` (304), then the tail.
4. Hand repair via #478.
5. Clear the 87 mirror findings (76 of them in `guides/quick-reference.md` × 4 locales).
   The checker already covers the mirrors as of PR #479; the normalizer does not.
6. **Flip `--warn` off** in the PR that clears the last batch. The red-to-green
   transition across these PRs is the evidence the gate can fail.

Tag-scoping keeps each diff reviewable and a bad batch revertable in isolation.

## Review notes for the batches

These are not rubber-stamp diffs:

- **Prose can reference in-fence identifiers.** `i18n/caveman-lite/skills/refactor-skill-structure/SKILL.md`
[...]

Issue #476, opened 2026-07-31
content: runaway and unclosed code fences in 11 English files mis-render ~2,242 lines across 71 files


pjt222:
## Defect

Eleven English content files nest three-backtick fences inside a ```` ```markdown ````
template. Under CommonMark the outer block closes at the first inner bare ```` ``` ````,
and the line intended as the *closer* then **opens** a new runaway fence that swallows
everything up to the next closer.

Worked example — `skills/escalate-issues/SKILL.md`:

| line | text | intended | actual |
|---|---|---|---|
| 92 | ```` ```markdown ```` | open template | opens |
| 112 | ```` ```language ```` | nested example inside template | body text |
| 114 | ```` ``` ```` | close nested example | **closes the template** |
| 136 | ```` ```text ```` | close template | **opens a runaway `text` fence** |
| 183 | ```` ``` ```` | — | closes it, 46 lines later |

Lines 137-182 therefore render as preformatted code: `**Expected:**`, `### Step 3:
Determine Routing`, and an entire ```` ```python ```` block. Confirmed with the shipped
extractor (`scripts/lib/fences.js`), which reports the `text` fence at L136 spanning 46
lines.

## Scope

Six SKILL.md: `escalate-issues`, `plan-release-cycle`, `setup-putior-ci`,
`tidy-project-structure`, `write-claude-md`, `write-vignette`. `setup-putior-ci`
has two runaways (`yaml` at L47 / 42 lines, `r` at L100 / 46 lines).

Five more files with the same or an adjacent defect: `agents/security-analyst.md`
(one fence unclosed to EOF at :243), `agents/senior-web-designer.md` (unclosed to EOF
at :125), `agents/senior-data-scientist.md`, `guides/agentskills-alignment.md`,
`guides/running-a-code-review.md`.

Roughly **2,242 lines mis-render as preformatted code across 71 files**, because the
60 translated mirrors reproduce the broken English faithfully.

## Why this blocks #472

It changes which lines parse as fences in those files. 18 of the 1,220 fence-parity
violations and the single `language`-tagged one sit inside runaway regions, so the
parity numbers must be re-measured after this lands. Left unfixed, ~62 phantom `text`
bodies stay pinned in the corpus permanently.

**Do not let byte parity drive this repair.** All 60 translated copies are faithful
mirrors of a broken English source; a parity-driven fix would rewrite the correct
files to match the broken one. Fix English first, then re-mirror.

## Fix

[...]

PR #518, opened 2026-08-10
feat(i18n): extend the fence normalizer to the mirrors, and restore all 87 (#477)

pjt222:
Clears the `agents`/`teams`/`guides` mirror slice of #477 — **the last mechanically-repairable part of the fence backlog.**

> ⚠️ **Conflicts with #512.** Both PRs edit `scripts/normalize-i18n-fences.js` and its test file. I tested the merge rather than assuming — it does not auto-resolve. Whichever lands first, I will rebase the other; I wrote both sides, so the resolution is mechanical. See the note at the bottom.

## The gap

The normalizer was skills-only while `check-i18n-fence-parity.js` covered all four content trees, so 87 of the 335 gated violations were flagged by one tool and unreachable by the other.

## The fix

Paths differ by tree — `skills/<id>/SKILL.md` against `<tree>/<id>.md` — and which names count as content is now decided by `contentKey` from `lib/fences.js`, the same function `buildEnglishFenceHistory` keys on.

That matters more than it looks. The history lookup was `history.get('skills/' + t.skill)` — a hand-built key that only *happened* to agree with `contentKey`'s output for the skills tree. Deriving both from one function is what stops `_template.md`, `README.md` and `_registry.yml` needing a second exclusion list here that could drift from the checker's.

Adds `--tree`, which scopes a run the way `--tag` does. Validated against the trees the repository **actually carries**, not against the `TREES` constant: a value naming a real tree the corpus has no translations for would otherwise produce the clean-looking zero `--locale` and `--tag` already guard against. `--tree teams` exits 2 here, because no team translation carries a divergence.

## The repair

```
normalize-i18n-fences.js --tree guides,agents,teams --write
  11 files, 87 fences    de=22  es=22  zh-CN=22  ja=21

  i18n/{de,es,ja,zh-CN}/guides/quick-reference.md   19 each
  i18n/{de,es,ja,zh-CN}/agents/r-developer.md        2 each
  i18n/{de,es,zh-CN}/agents/contemplative.md         1 each
```

All 87 are translated comments inside frozen fences — `# Die Skills zaehlen` back to `# Count the skills`, `# 线性混合效应模型` back to `# Linear mixed-effects model`. The diff is 341 lines each way and every one sits inside a gated fence: no prose, no headings, no frontmatter.

Restored from each translation's own `source_commit`, so nothing is entangled with the separate staleness backlog (#278).

## Measured

| | before | after |
|---|---:|---:|
| gated-fence violations | 335 | **248** |
| files | 75 | **64** |

Exactly the 87 fences and 11 files the checker attributed to the mirrors. The remaining 248 are the parked sets registered in #501 — the regulated carve-out, the #478/#483 ordinal-unsound files, and the #476-blocked mirrors. **Nothing mechanically repairable remains.**

## Tests

Five new, 29 total in the suite. Including that a guide's translated **prose and frontmatter survive** while only its frozen fence is restored — the failure mode that would matter most here — and that a `_template.md` dropped into a tree is not treated as content.

## On the conflict with #512

#512 adds the step-correspondence (fork) guard to the same file. This branch does **not** carry that guard, so these 87 fences were restored without it. That is acceptable for this slice and worth stating plainly: all 11 files pass the fence-count and tag-sequence soundness checks, and the diff was read — every change is a comment inside a fence, which is the shape a fork detector exists to distinguish *from*. Once both land, mirror batches get the guard automatically.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -272,9 +272,17 @@ fence-count and tag-sequence checks that decide whether ordinal mapping is
 trustworthy at all. A tag matching no divergent fence exits 2 rather than
 reporting a clean-looking zero.
 
-It repairs `skills/` only; the checker covers all four trees, so an agent, team
-or guide violation shows up as `files to change: 0` and needs repairing by hand
-(#477).
+`--tree <list>` scopes the same way across content trees, and the normalizer now
+covers all four — `skills`, `agents`, `teams`, `guides` — so it repairs exactly
+what the checker flags. It was skills-only until the mirrors became the last
+mechanically-repairable slice of #477 (87 fences / 11 files, cleared in #518).
+
+Both scoping flags are validated against what the run actually reached, never
+against a static list: a `--tag` matching no divergent fence exits 2, and so does
+a `--tree` naming a tree the selected `--locale` carries no translations for.
+Validating `--tree` corpus-wide instead let `--locale wenyan --tree guides` clear
+both guards independently and report `files to change: 0` — six of the ten
+locales carry `skills/` alone.
 
 Runs **warn-only** in CI until the backlog clears (#477), then flips to blocking.
 Warn is a temporary state with a named exit, not the design. Quote the current
--- a/i18n/README.md
+++ b/i18n/README.md
@@ -74,9 +74,11 @@ uncommitted work with it (#486).
 tag matching no divergent fence exits 2 rather than reporting zero, so a typo
 cannot read as "this batch is already done".
 
-It repairs `skills/` only; agent, team and guide violations are reported by the
-checker but show up here as `files to change: 0`, and need repairing by hand
-(#477).
+`--tree <list>` scopes the same way across content trees. The normalizer covers
+all four — `skills`, `agents`, `teams`, `guides` — so it repairs exactly what the
+checker flags; it was skills-only until the mirror slice was cleared in #518. A
+`--tree` naming a tree the selected `--locale` carries no translations for exits
+2, for the same reason a `--tag` matching nothing does.
 
 Runs **warn-only** in CI until the backlog clears (#477), then flips to blocking.
 
--- a/i18n/de/agents/contemplative.md
+++ b/i18n/de/agents/contemplative.md
@@ -134,10 +134,10 @@ Dieser Agent verwendet einen Kommunikationsstil der **stillen Praesenz**:
 ```yaml
 settings:
   depth: standard          # light, standard, deep
-  sequence: adaptive       # adaptive, fixed (meditate->heal->center->shine)
+  sequence: adaptive       # adaptive, fixed (meditate→heal→center→shine)
   expression: minimal      # minimal, moderate, full
   attunement: enabled      # enabled, disabled
-  memory_integration: true # Dauerhafte Einsichten in MEMORY.md schreiben
+  memory_integration: true # write durable insights to MEMORY.md
 ```
 
 ## Werkzeuganforderungen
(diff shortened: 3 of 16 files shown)


pjt222:
**Adversarial review: tool `correct`, repair `correct`, guards `defective`.** Copilot did not pick this up and the advocatus-diaboli agent has been idling, so this ran as a three-lens workflow review.

The **87-fence repair needed no change** and was verified far past its own claims. Both lenses independently re-derived it: 341 changed lines each way with **zero outside a gated fence**, every restored body byte-identical to the same-ordinal basis from that file's own `source_commit`, and — using a quote-aware stripper — every changed line differing only after an unquoted `#`, so the code skeleton is byte-identical. One lens went further and **extracted `measure()` from #512's branch**, running the fork guard this branch does not carry against the slice: 103 gated pairs, containment 1.00 throughout, 0 files refused. Restoring these unguarded was provably safe rather than assumed safe.

The defects were all in the new flag and its tests. Fixed in `86692423d` and `8cf140467`:

**`--tree` was validated corpus-wide, so composing it with `--locale` reproduced the exact clean-looking zero the guard family exists to reject.** `PRESENT_TREES` is a union over all locales and the check never intersected it with the selected locale, so on the real corpus:

```
$ node scripts/normalize-i18n-fences.js --locale wenyan --tree guides
exit 0    files to change: 0
```

Six of the ten locales carry `skills/` alone, so this is reachable with real names. Now validated **after** the scan against the trees the scoped run actually reached — the same shape as `--tag` — and the accept-list is collected before the `--tree` filter and after the existence checks, so it means "carries translated content" rather than "has a directory of that name":

```
$ node scripts/normalize-i18n-fences.js --locale wenyan --tree guides
ERROR: --tree matched no translated content in locale 'wenyan': guides
Reachable here: skills
exit 2
```

**The English dirty warning was still skills-only.** The basis is read off disk whenever a `source_commit` fails to resolve and always under `--basis head`, and `t.english` is now `guides/quick-reference.md` as readily as a `SKILL.md` — so an uncommitted edit there was spliced into four locales with no NOTE.

**A directory named `*.md` under a mirror tree killed the run with EISDIR**, where the checker skips it. For skills the entry is a directory and the file is `SKILL.md`, so existence alone was safe by construction; the mirror branch makes the entry itself the file.

**Two mutants survived all 29 tests**, and one was inside the test that named the property it failed to lock:

| mutation | was | now |
|---|---|---|
| delete `if (key === null) continue;` | 29 green — the template fell through to the unsound-mapping path and was still not rewritten, *while the report told a reviewer to hand-repair it* | KILLED by 1 |
| delete `if (!hasTree(locale, tree)) continue;` | 29 green — no fixture had a skills-only locale, the dominant real-corpus shape | KILLED by 2 |
| `--tree` post-scan check → no-op | (new guard) | KILLED by 2 |

Adding a skills-only locale to the fixture closed both gaps at once, since the `--locale`/`--tree` composition was unreachable without it.

**Three documentation sites** still said skills-only and the original diff touched none of them — `CLAUDE.md`, `i18n/README.md`, and worst, `skills/translate-content/SKILL.md`, an operational instruction telling translators to hand-repair the mirrors this PR automates. Found by grep rather than memory, which is the point.

29 → 32 tests. Corpus re-verified: `--tree guides,agents,teams` now plans 0 files because the slice is cleared, and the gate stands at 248.

One finding left open deliberately: `contentKey` applies its `_`/README exclusion only on the flat `<tree>/<id>.md` branch, so `contentKey('skills/_template/SKILL.md')` returns a key rather than null. Unreachable today (no locale carries a translated `_template`) and unchanged by this PR, but my commit message stated the exclusion as a general property of `contentKey`, which is wrong for the tree holding most of the corpus. Filing separately.

Issue #472, opened 2026-07-30
i18n: the keep-code-in-English rule is unenforced — 386 executable fences across 102 files carry translated content


pjt222:
## Gap

`CLAUDE.md` § *Translation Rules* says code blocks are kept in English. Nothing
checks it, and the corpus has drifted.

The three i18n validators — `check-i18n-frontmatter-parity.js`,
`check-translation-freshness.js`, `generate-translation-status.js` — compare
frontmatter, `source_commit` recency, and body-equality for stub detection. None
of them opens a fenced block. The rule is stated but unenforced.

## Measurement

Across all 10 locales, 3,576 translated `SKILL.md` compared against their English
sources, 22,554 fenced blocks:

| | fences | files |
|---|---|---|
| differ from the English counterpart | 2,366 (10.5%) | 661 (18.5%) |
| **hard violations** — CJK inside a fence whose English counterpart has none | **749** | **213** |
| of those, in an **executable-language** fence | **386** | **102** |
| of those, in `text` / `markdown` / untagged fences | 334 | — |
| unclassified (Latin-script locales) | 1,617 | 453 |

The CJK test is the discriminator that matters. Staleness cannot manufacture
Japanese characters: if a `bash` fence in `i18n/ja/` contains CJK and the English
fence contains none, someone translated inside a code block. Everything else is
confounded — a stale translation's fence matches an *older* English source and
diverges without anyone having touched it, so the 1,617 Latin-script divergences
mix real violations with mere lag and are **not** claimed here either way. With
2,545 translations already stale repo-wide, that confound is large.

Executable fences carrying translated content, by tag:

```
168  bash      113  r        53  yaml     28  python
  8  dockerfile   7  json      3  typescript   3  go
  1  javascript   1  sql       1  css
```

Samples:

```
[r]    i18n/ja/skills/add-puzzle-type/SKILL.md          # valid_typesに追加
[yaml] i18n/ja/skills/add-puzzle-type/SKILL.md          # タイプ固有のパラメータをここに追加
[json] i18n/ja/skills/build-grafana-dashboards/SKILL.md # ... (完全な設定はEXAMPLES.mdを参照)
```

Worst files by diverging fences, for a sense of the tail:

```
19/19  i18n/caveman/skills/test-a2a-interop/SKILL.md
19/19  i18n/caveman-lite/skills/test-a2a-interop/SKILL.md
17/18  i18n/ja/skills/enforce-policy-as-code/SKILL.md
[...]

Issue #481, opened 2026-07-31
i18n: retagging a frozen fence to `text` still evades the parity gate — scope is editable by the gated file


pjt222:
## Gap

The fence-parity gate (#472, PR #479) decides what to check from the info string of the
**translated** file. `isGated(fence)` reads `fence.lang` off the translation, and the
English history stores only `f.body` — the info string is never compared. So the set of
fences under the gate is chosen by the file being gated.

Retagging a ```` ```yaml ```` fence to ```` ```text ```` removes it from enforcement
entirely, and it is then absorbed into the reassuring summary line:

```
714 divergence(s) in ungated tags (text/markdown/untagged) — localising those is allowed.
```

Verified end to end: a control edit inside an ```` ```r ```` fence produces a blocking
finding; the same edit with the opener changed to ```` ```text ```` produces **0**.

## What default-deny did and did not fix

PR #479 inverted the polarity from an allowlist of code tags to "frozen unless `text`,
`markdown`, `md`". That **narrows** the escape set from unbounded to exactly three tags
— worth having — but does not close it. The claim in `scripts/lib/fences.js` that
default-deny stops scope being "editable by the person being gated" overstates; the
comment has been corrected to say so, and points here.

**0 live instances** across 3,427 count-matched pairs, so this is a hole not currently
being walked through.

## Fix

Compare the ordered info-string sequence between translation and English basis, and
flag any position where the tag changed. The comparison already exists in the sibling
tool — `scripts/normalize-i18n-fences.js` computes `alignmentTag` per position to
decide whether ordinal mapping is sound. Lifting it into the checker turns "is this
mapping safe?" into "did the tag change?".

Two knowns for the implementer:

- `text` facing untagged must stay benign. `normalize-content-style.js --mode fences`
  retro-tagged untagged blocks as `text` on one side only; `alignmentTag` folds them.
  Expressing it as `isGated(a) !== isGated(b)` is wrong under default-deny — untagged
  is gated, `text` is not — and that exact formulation stranded 169 repairable fences
  across 73 files mid-PR before it was caught.
- The measured population is **144 (locale, id) pairs** with a fence-count or
  tag-sequence mismatch. Ship warn-only until those clear, then blocking.

## Why it matters more than its zero count suggests

It is the difference between a gate whose scope is fixed and one whose scope is a
suggestion. #477 flips the parity gate to blocking; the moment it blocks, retagging
becomes the path of least resistance for anyone who wants a build to go green.

## Acceptance criteria

[...]

PR #520, opened 2026-08-10
feat(i18n): gate that a yaml fence is actually YAML, and fix the six it found (#507)

### LINKED ISSUE #507: i18n: hybrid fences mix machine-consumed keys with human-fillable prose, and a whole-fence freeze cannot govern both
## The question

#472's freeze operates on a whole fence: the tag decides, and everything between the delimiters is either frozen or localisable. Some fences carry **both** kinds of content, and neither answer is right for the whole thing.

Surfaced while scoping #477 batch 6 (`yaml`+`typescript`). Not a defect in that PR — the fence there is restored, correctly, per the settled rule.

## The example

`skills/escalate-issues/SKILL.md:193`, introduced by prose reading *"**For Specialist Agents** (structured format for MCP tools)"*:

```yaml
---
type: escalation
severity: high
from_agent: janitor
to_agent: security-analyst
blocking: false
---

# Security Concern: Hardcoded API Key in Config

**File**: config/production.yml:45
**Pattern**: API_KEY="sk_live_abc123..."

**Request**: Please review if this is a valid secret or a placeholder.
If valid, recommend secure credential management strategy.

**Context**: Discovered during config cleanup sweep.
```

The frontmatter is genuinely machine-consumed — `type`, `severity`, `from_agent`, `to_agent`, `blocking` are keys a tool parses, and the agent IDs are English by the ID rule. Freezing it is right.

Everything after the `---` is a report a human writes and another human reads. It is exactly what the `text`/`markdown` exemption exists to protect: per CLAUDE.md, those tags stay localisable because *"they carry reference tables, decision flows and report templates a human reads or fills in — a German reviewer should be able to emit a German report."*

Before batch 6, `i18n/de/skills/escalate-issues/SKILL.md` had that body in German. It is now English, and the surrounding German prose still instructs the reader to produce one.

## Why the obvious fixes are wrong

**Retagging to `text`.** CLAUDE.md names this explicitly as the attack the closed exemption list defends against: *"would let the scope be edited by retagging a ```yaml fence to ```text."* It would also unfreeze the frontmatter keys, which must stay English.

**An ad-hoc exemption for this file.** Same objection, less honest. Default-deny is a settled design decision.

**Leaving it divergent.** Not a fix; it keeps a parity violation and blocks #477 from dropping `--warn`.

## Options worth considering

1. **Split the fence in the English source.** A frozen ```yaml fence for the frontmatter, then a localisable ```markdown fence for the report body. Costs nothing structurally, needs re-mirroring to 10 locales, and makes the boundary explicit where a reader can see it.
2. **Lift the report body into prose**, the same remedy #502 proposes for comment-only fences. Consistent, and would let one pass address both shapes.
3. **Accept English-only for hybrid fences** and say so in the i18n rules, so it is a decision rather than an accident.

Option 1 looks best — it keeps the machine contract frozen, restores the human's ability to work in their own language, and is mechanically checkable afterwards.

## Scope

Unmeasured. This issue records one confirmed instance; the corpus should be swept for fences that contain a `---` frontmatter block followed by prose, and for frozen fences whose bodies are majority natural-language. That sweep overlaps #502 (196 comment-only frozen fences) and should probably be one triage pass over both shapes.

Tracked in the parked-set register: #501.

pjt222:
Closes #507.

#472 freezes a fence by its **tag**: everything between the delimiters is either frozen to English or localisable, and `yaml` is frozen. That is right for a machine-consumed document and wrong for a reference table a human reads — and the tag is the only thing telling them apart, so a mislabelled fence quietly freezes content that should have stayed translatable.

`skills/escalate-issues/SKILL.md` carried a `yaml` fence holding a routing header **and** the markdown incident report beneath it. The German translation had that report in German; #508 restored it to English, correctly per the rule, and the surrounding German prose still told the reader to write one.

## What makes it decidable

Copilot's observation on #508, recorded in #507: that fence **is not YAML**. `js-yaml` rejects it at the `---` separating header from prose. So the rule is mechanical — if it is tagged `yaml`, it parses. If it does not, it is either broken YAML or it is not YAML, and both want fixing at the source rather than freezing into ten locales.

This is also what keeps retagging honest. CLAUDE.md names retagging `yaml` → `text` as the way the freeze's scope could be edited from inside the gated file. Correcting a mislabel is not that — but the two are only distinguishable if something checks which fences are genuinely YAML. Now something does.

## The gate

`scripts/check-yaml-fences.js`, blocking in CI, repo-wide rather than added-lines-only because the corpus is at zero and there is no legacy backlog to grandfather.

**Exemptions are by error class, not by location.** Two shapes legitimately do not parse — Helm Go templates, and duplicate-key illustrations where a good/bad pair in one fence is a documented idiom in these guides. Pinning exemptions to `file:line` would rot on the first edit above them, and a rotted exemption either fails a clean file or silently covers a new defect. Keying on the error class costs precision (an accidental duplicate key in a real config would pass) and buys a list that cannot drift. **Every exempted fence is reported, not skipped** — an exemption list nobody can see is one nobody notices growing.

## What the first run found

331 English yaml fences, 15 non-parsing: 5 Go templates and 5 duplicate-key illustrations (both exempted), plus **six things to fix** — one more than #507 recorded, which had listed `render-publication-graphic` as unclassified.

| | fix |
|---|---|
| `escalate-issues:194` | **split** — frozen `yaml` header + localisable `text` report, with prose naming which is which |
| `troubleshoot-print-issues` ×3 | **retagged `text`** — ranges (`1.0-2.0mm`), arrows (`0.98 → 0.96`), bullet lists after a mapping |
| `render-publication-graphic:51` | **retagged `text`** — publication requirements with `RGB or CMYK (check guidelines)` |
| `write-helm-chart:328` | **a real defect** — an elision comment dropped into the middle of a mapping, orphaning a `tls:` block under nothing. Same shape as #476's runaway fences. Its two values-file excerpts now carry a `---` separator, so the fence parses as the two documents it always was rather than riding the duplicate-key exemption |

## Re-mirrored into 30 files across 10 locales

**The English retag alone is half a fix.** The parity gate reads a fence's tag off the *translated* file, never off English, so retagging English would have left every mirror frozen — exactly the content this frees — while diverging the tag sequence and making the normalizer refuse those files. I checked this rather than assuming it.

Every mirror carried the English body verbatim (measured 10 of 10, none structurally diverged), so each fence was addressed on its own. A whole-region diff was tried first and matched in **no** locale: the retagged fences are far apart, so the changed region swallows the translated prose between them.

Tag sequences afterwards: `troubleshoot-print-issues` 10/10 aligned, `render-publication-graphic` 10/10, `escalate-issues` 7/10. The three stragglers are **pre-existing and unrelated** — `wenyan`, `wenyan-lite` and `zh-CN` tag fence 3 `text` where English has `python`, and they were already diverged before this change. Filing separately.

## Verification

```
gate:   331 checked, 9 exempted, 0 failures
parity: 248 violations — unchanged, as expected
```

Parity is unchanged because these fences carried English and were never violations. What changes is that a translator may now localise them.

**Four mutants, all killed:**

```
failures.push -> exempted.push          (the gate can fail)     KILLED by 4
delete the go-template exemption                                KILLED by 2
delete the duplicate-key exemption                              KILLED by 1
loadAll -> load                          (multi-document)       KILLED by 1
```

12 tests, each fixture the shape of something the first corpus run actually found — including the paired positive that the prescribed remedy, retagging to `text`, makes the same body pass.

### DIFF
--- a/.github/workflows/validate-content-style.yml
+++ b/.github/workflows/validate-content-style.yml
@@ -10,6 +10,7 @@ on:
       - 'guides/**'
       - 'i18n/**'
       - 'scripts/check-yaml-fences.js'
+      - '.github/workflows/validate-content-style.yml'
   pull_request:
     paths:
       - 'skills/**'
@@ -18,6 +19,7 @@ on:
       - 'guides/**'
       - 'i18n/**'
       - 'scripts/check-yaml-fences.js'
+      - '.github/workflows/validate-content-style.yml'
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/validate-content-style.yml
+++ b/.github/workflows/validate-content-style.yml
@@ -3,9 +3,21 @@ name: Validate Content Style
 on:
   push:
     branches: [main]
-    paths: ['skills/**', 'agents/**', 'teams/**', 'guides/**', 'i18n/**']
+    paths:
+      - 'skills/**'
+      - 'agents/**'
+      - 'teams/**'
+      - 'guides/**'
+      - 'i18n/**'
+      - 'scripts/check-yaml-fences.js'
   pull_request:
-    paths: ['skills/**', 'agents/**', 'teams/**', 'guides/**', 'i18n/**']
+    paths:
+      - 'skills/**'
+      - 'agents/**'
+      - 'teams/**'
+      - 'guides/**'
+      - 'i18n/**'
+      - 'scripts/check-yaml-fences.js'
   workflow_dispatch:
 
 permissions:
@@ -36,3 +48,19 @@ jobs:
       - name: Repo-wide scan (informational)
         if: github.event_name != 'pull_request'
         run: node scripts/check-content-style.js --all
+
+  # Blocking, and repo-wide rather than added-lines-only: the corpus is at zero
+  # (#507), so there is no legacy backlog to grandfather, and a fence tagged
+  # `yaml` that is not YAML freezes content into ten locales that should have
+  # stayed translatable. Its own job because it needs `npm ci` for js-yaml,
+  # which the style job does not.
+  yaml-fences:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-node@v7
+        with:
+          node-version: '24'
+          cache: 'npm'
+      - run: npm ci
+      - run: npm run validate:yaml-fences
--- a/i18n/caveman-lite/skills/escalate-issues/SKILL.md
+++ b/i18n/caveman-lite/skills/escalate-issues/SKILL.md
@@ -196,6 +196,10 @@ def route_issue(severity, issue_type):
 Generate a formatted report suitable for the target audience (agent or human).
 
 **For Specialist Agents** (structured format for MCP tools):
+Emit one document: the routing header delimited by `---` fences, then the report
+body beneath it. The receiving tool parses the header keys, so those and the
+agent ids stay in English.
+
 ```yaml
 type: escalation
 severity: high
@@ -204,8 +208,6 @@ to_agent: security-analyst
 blocking: false
 ```
 
-The body beneath it is a report a human writes and another human reads, so it is
-localisable and a German reviewer should be able to emit a German one:
 
 ```text
 # Security Concern: Hardcoded API Key in Config
--- a/i18n/caveman-lite/skills/escalate-issues/SKILL.md
+++ b/i18n/caveman-lite/skills/escalate-issues/SKILL.md
@@ -197,14 +197,17 @@ Generate a formatted report suitable for the target audience (agent or human).
 
 **For Specialist Agents** (structured format for MCP tools):
 ```yaml
----
 type: escalation
 severity: high
 from_agent: janitor
 to_agent: security-analyst
 blocking: false
----
+```
 
+The body beneath it is a report a human writes and another human reads, so it is
+localisable and a German reviewer should be able to emit a German one:
+
+```text
 # Security Concern: Hardcoded API Key in Config
 
 **File**: config/production.yml:45
--- a/i18n/caveman-lite/skills/render-publication-graphic/SKILL.md
+++ b/i18n/caveman-lite/skills/render-publication-graphic/SKILL.md
@@ -53,7 +53,7 @@ Produce publication-ready graphics that meet technical requirements for academic
 
 Identify technical specifications for target publication:
 
-```yaml
+```text
 # Common publication requirements
 
 academic_journal:
--- a/i18n/caveman-lite/skills/troubleshoot-print-issues/SKILL.md
+++ b/i18n/caveman-lite/skills/troubleshoot-print-issues/SKILL.md
@@ -196,7 +196,7 @@ Implement immediate solutions for common issues:
 ```
 
 **Retraction tuning**:
-```yaml
+```text
 # Direct drive extruder:
 retraction_distance: 1.0-2.0mm
 retraction_speed: 40-50mm/s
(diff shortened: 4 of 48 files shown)


pjt222:
**Adversarial review: all three lenses `defective`.** Fixed in `c23896728`.

The mechanical claims all held under independent re-derivation — 331/9/0, tag sequences 10/10 · 10/10 · 7/10 with the three stragglers confirmed pre-existing at `HEAD~2`, and parity 248 before **and** after with a byte-identical violation set, so no pre-existing violation was hidden by moving 40 fences out of the gated set. That was the real risk of a mass retag and it was checked properly.

Four defects sat behind those numbers. Three are mine to own.

### The exemption was keyed on the body, not the error

`/\{\{.*\}\}/s` reads as "is this a template?" and means "does this fence contain two braces anywhere?" It matched **24 of 331** English fences where only **4** are Helm sources. The other 20 are valid GitHub Actions (`${{ matrix.config.os }}`) and Prometheus documents — the *most* machine-consumed fences in the corpus — and for all 24, any parse error whatsoever was forgiven.

The reviewer demonstrated it with two fences differing only by a `${{ }}` expression, both carrying the exact orphaned-`tls:` defect this gate was built to catch: one reported exempted, one failing.

Now keyed on the error's own line. All four genuine Helm errors land on a `{{ … }}` line; a broken workflow's does not. This also makes the header's "exemptions are by ERROR CLASS" claim **true**, which it was not — an untested assertion of exactly the kind this repo has a rule about.

### English-only scope let this PR's own headline fix escape

The orphaned-`tls:` repair reached English and **none of its ten mirrors**, all of which still failed to parse. Nothing could ever have reported them: the parity gate accepts a translated body matching any historical English revision, and a stale broken body is precisely that.

The gate now walks the mirrors too — **2502 fences** — and its first run named all ten. Mirrored, and the gate is now green corpus-wide.

### The re-mirror inverted this PR's own thesis

It injected an untranslated rationale line — *"a German reviewer should be able to emit a German one"* — verbatim into Japanese and classical-Chinese files, while **deleting the `---` document contract from every locale** and adding its replacement to English only. A tag-driven edit silently removing content from translated files is the thing this PR exists to prevent.

The rationale is now gone from all 11 files (it is toolchain policy, not skill content — it already lives in the script header), and the contract sentence is in all 11.

### Also

`--root` was validated by `existsSync` + `isDirectory` — a proxy for "would this scan anything", so an empty directory was a clean run. It now errors when no content tree is found. And js-yaml 5's schema rejects CloudFormation `!Ref` and even core `!!binary`, which are genuinely machine-consumed; those get their own exemption class rather than a failure message recommending the retag CLAUDE.md warns against. The message now leads with *fix the YAML* and names retagging as the exception.

### Tests

12 → 19, including the two the review proved were missing:

```
body-keyed -> error-line-keyed exemption      KILLED by 1
path-keyed exemption ("helm" in the path)     KILLED by 1
mirrors dropped from scope                    KILLED by 1
empty-root guard disabled                     KILLED by 1
  ...plus the original four.
```

Two of my new fixtures were wrong on first write — one asserted a message that no longer existed, the other **parsed cleanly and so tested nothing**. Both fixed rather than deleted.

### Left as-is, deliberately

The `ingress.hosts: [my-app.example.com]` shorthand the reviewer flagged is a list of strings where the chart's own template wants mappings — but that shorthand already exists on the adjacent unchanged `values-dev.yaml` line and `SKILL.md:433` refers to it as a teaching example, so this propagates a pre-existing simplification inside a block marked `(excerpt)` rather than originating one. Filing with the EXAMPLES.md pointer the same reviewer noted was dropped.

Two low-severity items also filed rather than widened into this PR: the `_template` coverage asymmetry in `contentKey`, and the absence of a mirror-vs-English tag-sequence check (#481 is unshipped, so the 10/10 verification here was manual).

Issue #632, opened 2026-08-17
skills: two more Deployment-versus-HPA conflicts, one of them the raw-manifest form #551 did not cover


pjt222:
#551's closing note suggested grepping the Kubernetes-adjacent skills for the same pattern while fixing it. Done, and it found two live instances plus one clean file. Filed rather than fixed, to keep the #551 PR reviewable.

The grep also turned up a **second form** of the defect that #551's framing does not describe.

## The form #551 describes — Helm values

`skills/deploy-to-kubernetes/references/EXAMPLES.md`

```yaml
replicaCount: 3          # :489
...
autoscaling:             # :525
  enabled: true
  minReplicas: 3
```

Same values.yaml, and the `helm create` scaffold template this file is teaching guards `replicas:` behind `{{- if not .Values.autoscaling.enabled }}`. So `replicaCount: 3` is inert exactly as in #551. Milder here only because `3` happens to equal `minReplicas: 3`, which hides the contradiction rather than removing it.

## The form #551 does NOT describe — raw manifests

A Deployment that declares `replicas:` **and** an HPA targeting that same Deployment. No Helm, no template guard, nothing inert: the two controllers fight, and every `kubectl apply` resets the replica count the autoscaler just chose. This is the standard footgun the `{{- if not ... }}` guard exists to prevent, and both skills teach it without comment.

**`deploy-to-kubernetes`**

| what | where |
|---|---|
| `Deployment/myapp` (ns `myapp-prod`), `replicas: 3` | `SKILL.md:231`, `references/EXAMPLES.md:175` |
| HPA `scaleTargetRef` → `Deployment/myapp`, `minReplicas: 3` | `references/EXAMPLES.md:392-402` |

**`deploy-ml-model-serving`**

| what | where |
|---|---|
| `Deployment/churn-prediction` (ns `seldon`), `replicas: 3` | `references/EXAMPLES.md:209-215` |
| HPA `scaleTargetRef` → `Deployment/churn-prediction`, `minReplicas: 2` | `references/EXAMPLES.md:658-668` |

Here the numbers openly disagree — `replicas: 3` against `minReplicas: 2` — so a reader cannot tell which is the intended floor.

**`configure-api-gateway`: clean.** It sets `replicas:` in several manifests and defines no HorizontalPodAutoscaler anywhere, so there is no conflict. Recorded so the next pass does not re-grep it.

## Blast radius

[...]

PR #841, opened 2026-09-15
docs-only: session log of the lead/support two-session pattern

pjt222:
Adds `docs/investigations/lead-support-coordination-2026-09-15.md`, beside the other investigation records. One new file, no other change.

Two interactive Claude Code sessions worked Wave 2 of the backlog plan on one checkout today: an Opus lead holding the plan, the branches and the merges, and a Sonnet peer session — a separate session, not a spawned subagent — doing the read-heavy scouting and two ten-file frozen-fence propagations. The operator asked both to file a session log on the pattern alongside the work.

## Shape

- **§1** the lead's account: what was delegated, what it bought, what it cost.
- **§2** the support session's own words, unedited by the lead — including two things only it could see.
- **§3** joint findings, drafted from both sections and sent to the support session for correction before this commit.

The assembly went through the lead only because the two sessions share one worktree and one branch. That is finding §3.3, not an editorial choice.

## Why it is worth committing

The arrangement is easy to describe as working well and hard to describe accurately. Every claim is a command's output or a transcript count (10 messages, 4,018 words; 67 tool calls on the support side; two propagations, zero refusals), and the failure modes each session could see from its own side get more room than the successes.

Three findings generalise past this pair:

**§3.1 — a brief inherits authority it has not earned.** Twice, the plan's shorthand for a state it *intended to create* reached the support session as an existing fact to locate, sourced to "the issue" — once for a stance framing that exists nowhere verbatim, once for five line numbers #632 never mentions. Neither session caught either at the point where catching it would have mattered.

**§3.4 — a claim about who found what needs the same discipline as a line number.** An earlier draft of §3.1 credited the support session with catching the second misattribution. It had not: it wrote that from recollection into its own §2, the lead found it plausible *because it flattered the support session and cost the lead*, and promoted it to established fact. A claim neither session had verified passed through both, each treating the other's account as its source — in a document whose subject is verification discipline. The underlying fact was true; the attribution was invented and survived two readings because it sounded like something that would have happened.

**§3.2 — `rg` exits 1 on no match**, so a chain of checks joined with `&&` stops at the first absent pattern and reports nothing about the rest. No error, just shorter output that still looks complete. It is not `rg`-specific — the same shape recurs with any command whose "found nothing" is a non-error non-zero exit. The support session hit it twice and caught it only because a count looked short, and notes that such a near-miss never reaches a lead: it arrives as a correct final report.

Also recorded: the read-only-then-named-paths-then-refusing-tool progression for write scope (§3.3), what the division of labour actually was as opposed to how it is usually described (§3.6), where the arithmetic turns in favour of delegating (§3.7), and two things this session did **not** test — a peer session against a spawned subagent for the same work, and a support session that disagrees with a *conclusion* rather than with a brief's framing.

## Checks

```
npm run validate:line-endings                OK: no committed CRLF in tracked text files
npm run validate:content-style -- --added    the new file is not flagged
wc -l -w                                     273 lines, 3,661 words
```

No registry, skill, agent, team or i18n file is touched, so no propagation, scaffold or README regeneration applies.


pjt222:
**docs-only** — no agent review round on this one, per the standing rule: documentation-only PRs go to the maintainer, and the Fable budget stays on PRs that change behaviour.

What that means for reading it: nothing here is checked by anything but the two gates below. The claims are of three kinds and they are not equally hard —

- **Counted from a transcript** (10 messages, 4,018 words, 67 tool calls on the support side, two propagations, zero refusals). Re-derivable, and the commands are named in the document.
- **Verified live by both sessions** — the two misattributions in §3.1, each checked independently on each side after one of us got it wrong.
- **Judgement** — §3.5's claim that the mirror trade-off came out better, and §3.7's claim about where delegation stops paying. Both are labelled as judgement in the text rather than dressed as measurements.

The one I would push on if I were reviewing it: §3.7 says the arrangement pays on breadth and not on single lookups, and that boundary is drawn from ten messages in one session. It is the weakest generalisation in the document and it is stated more confidently than that sample supports.

PR #843, opened 2026-09-15
fix(skills): an open comment wins over the fence rule in the budget estimator (#734)

### LINKED ISSUE #734: [MED] manage-memory: the fence and comment rules collide, and the stale-comment cascade drops real index lines
## Summary

`skills/manage-memory` ships a budget estimator whose fence/comment state machine decides what counts toward the auto-memory caps. #721 (PR #725) measured it correct on the **info-string** axis — five variants including untagged, all preserved-and-counted — and on the **position** axis, where a new `midbare` arm showed a bare comment is stripped wherever it sits.

Neither axis is where the detector is most likely wrong. The two rules **collide**, and several CommonMark cases diverge from what the harness does. All of the below are detector-vs-harness divergences whose harness-side truth is unmeasured.

## Findings

1. **[MEDIUM]** **The rules collide, and the result cascades.** The ` ``` ` branch fires *before* the comment branch:

   ```python
   if ln.lstrip().startswith('```'):
       fence = not fence
   elif not fence and (cmt or ln.lstrip().startswith('<!--')):
       cmt = '-->' not in ln
       continue
   ```

   A fence delimiter **inside an active bare comment** toggles `fence` and is itself kept. The rest of the comment is then kept as "fenced", `cmt` stays `True`, and when a later ` ``` ` flips `fence` off, the stale `cmt` resumes stripping **ordinary index lines** until one contains `-->`. A commented-out code block in a real index cascades into silently dropping unrelated entries. This needs no CommonMark subtlety — it is the detector's own two rules disagreeing.

2. **[LOW]** **Indented fences.** A ` ``` ` indented 4+ spaces is an indented *code block* in CommonMark, not a fence. The detector's `lstrip()` toggles on it.

3. **[LOW]** **Info string inside an open fence.** A ` ```yaml ` line *inside* an open fence is content in CommonMark; the detector toggles on it. Realistic in a `MEMORY.md` that documents fences — as this repo's own does.

4. **[LOW]** **Fence-length matching.** CommonMark closes a fence only with a run at least as long as the opener. The detector toggles on any ` ``` `, so a ````` ```` ````` block containing ` ``` ` closes early — the same swallowed-opener shape #628 records elsewhere.

5. **[LOW]** Not covered by the #721 arms either, and named in that run's NOT-claims: `~~~` fences, a tag with attributes (` ```js {highlight} `), and an unclosed fence.

## Acceptance Criteria

- [ ] Finding 1 reproduced against the harness with a wire-capture arm (a bare comment containing a fence delimiter), so the divergence is measured rather than reasoned
- [ ] The detector's state machine either handles the collision, or documents the case with the measurement behind it
- [ ] Findings 2–4 each get an arm or an explicit "unmeasured, and here is the risk direction" line
- [ ] Any case where the detector **under-reports** usage relative to the harness is fixed rather than documented — that is the dangerous direction this corpus names
- [ ] `skills/verify-memory-integrity` is checked for the same block, since the two ship the same logic

## Context

Found by an adversarial review of PR #725 on 2026-08-25. The instrument for all of this now exists and is cheap: `tools/wirecap.py` plus the arm generator in `tests/results/2026-08-25-fence-strip-replication/`.

Related: #721, #407, #628.

pjt222:
Closes #734 findings 1 and 5; documents findings 2 to 4 with their risk directions, per that issue's own acceptance criteria.

## Premises

New section, and the reason for it is in the review history of #839 and #842: both were blocked by a premise nobody had written down. Each line names what this fix depends on and where it was checked — attack these first.

| This fix is correct only if… | Verified where | Result |
|---|---|---|
| the estimator block is byte-identical in both carriers, and something enforces that | `scripts/test/memory-blocks.test.js:33-45` | Holds. `verify-memory-integrity` is `SOURCE_OF_TRUTH`; an **absent** copy fails as loudly as a divergent one |
| the collision drops *real index lines*, not just comment content | traced, then pinned by a new arm with a fixture whose `-->` sits inside the fenced region | Holds: repaired block keeps 12 lines, colliding one keeps 5 |
| repairing toward CommonMark cannot make things worse | **unverifiable here** — the loader's own behaviour is uncaptured | Only finding 1 is repaired, on the narrower ground that the block is self-inconsistent and drops lines. Findings 2 to 4 are documented with risk directions instead |
| `manage-memory`'s 10 mirrors need the fence propagated | `npm run check:fence-propagation -- --id manage-memory` | **False.** All ten are `UNALIGNABLE` (folded tag sequence differs from English) and none carries the block. The plan called this "translated mirrors → debt"; the real reason is structural and pre-existing |
| `verify-memory-integrity`'s 4 mirrors carry it | `rg` over the four, then `refresh:stubs` | Holds — all four are scaffold stubs, so propagation is a tool run |
| the 500-line cap on the English `SKILL.md` is the cap that binds | `wc -l` on the English file, against `.github/workflows/validate-skills.yml` | **False, and it cost a red required check.** The gate runs over `i18n/*/skills/*/SKILL.md` too, and a refreshed stub is English's body plus six provenance frontmatter fields — so the binding ceiling for English is **494**. English landed at exactly 500 and passed locally while all four mirrors failed CI at 506. Added as a row after the fact: the premise was checked in the file being edited rather than in the file the gate reads |

## The defect

The block's two rules disagreed, and the disagreement dropped real index lines.

The fence branch ran first, so a ``` delimiter inside an open block comment toggled `fence` and was itself kept. `cmt` then stayed set across the fenced region — the `elif not fence and …` guard cannot clear it while the fence is open — and when a later ``` closed the fence, the stale `cmt` resumed stripping **ordinary index lines** until a line containing `-->` appeared, which may never happen. The estimator then reports fewer lines than the loader will read: **under-reporting**, which hides a truncation that is already happening, and which #734's acceptance criteria single out as the direction that must be fixed rather than documented.

## The repair

Precedence. An open comment is tested first, because a ``` run inside a comment is comment content and not a delimiter — which is also what CommonMark says. No path can now carry `cmt` across a fence toggle. Both carriers change in the same commit.

## The arm, and proof it can fail

`scripts/test/memory-blocks.test.js` gains a behavioural arm. The fixture is shaped so that `cmt` survives: the comment's `-->` sits **inside** the fenced region, with ten real index entries after it.

```
npm run mutation-check -- --file skills/verify-memory-integrity/SKILL.md \
  --replace 'if cmt::if 0 and cmt' --test 'node --test scripts/test/memory-blocks.test.js'

[1/5] baseline (expect green) ... green (22 passing).
[4/5] running tests against the mutant (expect red) ... exit 1, 4 failing
MUTANT KILLED by 4 failing test(s) — the check can fail.
```

Four, not one: mutating the source of truth also desynchronises the byte-identity gate, which is by design — that gate fails an absent or divergent copy. Reasoned, not separately measured: the behavioural arm is one of the four.

## Findings 2 to 4

Documented rather than repaired, in `references/EXAMPLES.md` beside the strip section, each with **the direction it errs** — since only under-reporting is dangerous. The reason for documenting is stated there: the correct target is the loader's behaviour, and a "fix" toward CommonMark could widen the divergence it means to close. The arm that would settle any of them is `tools/wirecap.py` plus the generator in `tests/results/2026-08-25-fence-strip-replication/` — a capture, not an argument, and it is the operator's to run.

## The 500-line cap

Twice, and the second time is the interesting one.

The divergence table first landed in `SKILL.md` and took it to **516 lines against a cap of 500** (`validate-skills.yml:75`). `npm run validate:skill-sections` does not check the count, so the local suite stayed green while a required CI context would have gone red. Worth naming: that is a gate with no local equivalent. The table moved to `references/EXAMPLES.md` — the repository's documented progressive-disclosure pattern — and the file came to exactly 500.

Exactly 500 passed locally and **failed CI on two contexts, one of them required**. The same gate walks `i18n/*/skills/*/SKILL.md`, and a refreshed stub is the English body plus six provenance frontmatter fields, so all four mirrors landed at 506. The ceiling that binds English here is **494**, not 500, and this file sat at 493 before any of this — one line of headroom — so any edit to it forces an extraction.

`CLAUDE.md` § Skill Validation carried the narrow check — `skills/*/SKILL.md` only — so the documented local pre-check could not have caught this. It now walks both trees with one `find` predicate, and states that the ceiling binding an English file with scaffolded mirrors is 494. That is the ninth file in this PR.

The first version of that snippet was itself wrong, and the correction is in `525caab34`: I carried "`-gt`, never `-le`" across from #842 without re-deriving its precondition. #842's shape is `if [ … -le … ]; then refuse; fi`, where a non-integer operand makes `[` exit 2, `if` reads nonzero as false, and the refusal in the `then` arm is skipped. The snippet's shape is `[ "$n" -gt 500 ] && echo OVER`, which has no arm to fall into — `&&` short-circuits and the file passes in silence. **Fail-open, under a comment claiming fail-closed.** The operator was never the mechanism; where the refusal sits is. Measured on bash 5.2.21 and zsh 5.9, both operand positions, and independently by a peer session. Guarded now with a `case "$n" in ''|*[!0-9]*)` test ahead of the comparison and a `REFUSED` branch when `find` matches nothing, both exercised in both directions.

The remedy for the cap itself is the one `validate-skills.yml` names in its own comment: extract to `references/EXAMPLES.md`, and keep `OVERRUN_ALLOWLIST` empty. Step 1's measurement *provenance* moved — the versions the UTF-16 behaviour was measured on, the upstream issue, the reason to treat it as version-volatile. Every operative statement stayed in `SKILL.md`: what the cap counts, which direction a byte count and a character count each err, and the strip's two halves. English is now **491**, the mirrors **497**, and a scan of every `SKILL.md` and every mirror reports nothing over 500.

## i18n

`verify-memory-integrity`'s four mirrors carry the scaffold literal, so they are refreshed and stamped mechanically. `manage-memory`'s ten mirrors are declared debt for the structural reason in the Premises table, not merely because they are translated.

## Checks

```
node --test scripts/test/memory-blocks.test.js          22 pass, 0 fail
npm run mutation-check (above)                          MUTANT KILLED by 4
npm run validate:skill-sections                         0 skill(s) reported
npm run validate:untagged-fences                        0 untagged openers across 560 files
npm run refresh:stubs -- skills verify-memory-integrity --verify
                                                        4 stub(s), 0 diverged, 0 without fence_basis_commit
wc -l skills/verify-memory-integrity/SKILL.md           491 (English cap 500; binding cap 494)
wc -l i18n/{de,es,ja,zh-CN}/skills/verify-memory-integrity/SKILL.md
                                                        497 each (cap 500)
wc -l skills/manage-memory/SKILL.md                     331
every skills/*/SKILL.md and i18n/*/skills/*/SKILL.md    none over 500
```


pjt222:
## Handover — no agent review round on this one

The Fable reviewer ran out of credits partway through #842, so #843 has had **no adversarial round at all**. Everything below is self-verified. Naming the weak claims myself, since nobody else has attacked them.

**Green:** 17/17 at `525caab34`. Every number in the body re-derived against this head rather than carried: 22 pass / 0 fail, `MUTANT KILLED by 4` (re-run *after* the prose edits, because the file changed under the original proof), 0 skills reported, 0 untagged fences across 560 files, 4 stubs / 0 diverged, 491 English / 497 each mirror, and a corpus scan over every `SKILL.md` and every mirror reporting nothing over 500.

### The three weakest claims, in order

1. **"Repairing toward CommonMark cannot make things worse" is unverifiable from here, and the PR says so.** The loader's actual fence/comment behaviour has never been captured. That is why only finding 1 is repaired — on the narrow, measurable ground that the block is self-inconsistent and drops real index lines — and findings 2 to 4 are documented with their risk directions instead of fixed. The arm that would settle any of them is `tools/wirecap.py` against a live session, which is yours to run, not mine to simulate.

2. **The four documented divergences carry reasoned risk directions, not measured ones.** "Under-reports" versus "either direction" comes from reading the state machine against CommonMark, not from observing the loader. If the loader diverges from CommonMark in the same place, a direction could be wrong. They are recorded rather than acted on for exactly this reason.

3. **The 494 figure is empirical, not structural.** It is `500 − 6`, where 6 is the provenance frontmatter on the four current stubs (`locale`, `source_locale`, `source_commit`, `fence_basis_commit`, `translator`, `translation_date`). A mirror with a different frontmatter shape — a hand-translated file, or a future field — shifts it. The check now committed to `CLAUDE.md` walks both trees and does not depend on the number; the number is in the prose as an explanation, and it is the part that can go stale.

### What is solid

The defect and its repair are measured, not argued: the fixture puts `-->` inside the fenced region with ten real entries after it, the repaired block keeps 12 lines where the colliding one keeps 5, and the mutant that reverts the precedence is killed by 4 tests. The `manage-memory` premise in the Premises table was **false** and that is recorded rather than quietly dropped — all ten of its mirrors are `UNALIGNABLE` and none carries the block, so the plan's stated reason for the debt was wrong too.

### One process note

The `## Premises` mechanism this PR introduces **failed on its own first outing**. The table had five rows and none of them asked what the CI gate measures — which is exactly how the cap failure got through. A row for "what is the gate's scope, read from the gate" is now mandatory whenever a fix touches a file a gate measures. The table catches premises you think of; it does not generate them.

pjt222:
## Adversarial round ran (Opus); the estimator survived, its prose did not. Fixed at `d5f08a487`

The handover said this PR had had no adversarial round at all. It has now — Fable is out until Sunday, so the reviewer ran on Opus rather than being skipped. It read the bundle cut at `525caab34` (`BUNDLE_STATUS` clean).

**Verdict on the code: it survives.** The reviewer extracted the estimator from both carriers, rebuilt the pre-#734 loop from this PR's own diff hunk, and ran both against 25 fixtures. The repair is correct everywhere the two differ, the carriers are byte-identical (`cmp` silent, sha256 matching, 2,533 bytes), and each of the four mirrors differs from English by exactly the six provenance lines. Nothing in the estimator blocks.

It also supplied a sharper instrument than the PR body used. This PR proved its new arm with `--replace 'if cmt::if 0 and cmt'`, which kills the comment machinery wholesale and dies to **4** tests. Running the *actual* pre-fix loop instead dies to **exactly 1** — the new arm, 12 expected against 5 observed. That 1 is the meaningful number; per this repository's own "15 versus 1" warning, the 4 should not have been quoted as a measure of coverage strength, and I am retracting it here.

Everything below was **re-derived independently** before being acted on.

### Two checks the reviewer asked for before merge — both dissolve

**F1: does any `manage-memory` mirror carry the estimator block?** The premise row said "all ten are UNALIGNABLE … and none carries the block", and the reviewer correctly noted that `UNALIGNABLE` is the tool refusing to answer, not the answer — so the second clause was uncited. If a mirror carried the old colliding block it would ship this exact defect, and no gate would see it.

```
scanned 10 mirror(s), 0 carry the block
```

Premise confirmed. **My first attempt at this scan reported `mirrors scanned: 0`** — a git pathspec that matched nothing — and would have printed a clean-looking confirmation from an empty set. It is only trustworthy because the count is in the output.

**F7: do the mirrors' provenance shas name bytes they actually carry?** All four stamp `20fe5e96f`, while the PR head is later.

```
read at stamp: 28015 bytes
read at head : 28015 bytes
VERDICT: IDENTICAL
```

### Three findings fixed — all of them in prose this PR added

**F2 — the corpus half of the cap snippet had neither of the two guards the prose claims for it.** Both live in the per-ID half. Measured, block copied verbatim:

```
A: neither tree present      → "cap scan complete — no OVER line above means nothing is over"  exit=0
B: skills/ only, i18n/ gone  → "OVER: skills/demo/SKILL.md (501 lines)"
                               "cap scan complete — no OVER line above means nothing is over"  exit=0
```

B is the bad one for this PR specifically: the entire point of the change is that the mirrors are the tree that binds, and the scan announces an all-clear having skipped that tree. `find` complains to stderr, so a reader following the snippet's own instruction to read the last line sees a pass. Now it REFUSES an empty scan and puts the count *inside* the reassuring line:

```
A → REFUSED: no SKILL.md found — run this from the repository root
B → cap scan complete: 1 file(s), no OVER line above means nothing is over
```

`1 file(s)` where the corpus holds hundreds refutes its own sentence, which is the property a bare all-clear cannot have.

**F6 — `ID=<skill-name>` is a syntax error as printed.** `<` and `>` are redirection operators, so the placeholder terminates the assignment word:

```
bash  exit=2  bash: line 1: syntax error near unexpected token `newline'
zsh   exit=1  zsh: parse error near `\n'
```

Nothing below it runs either, when the block is pasted whole. `ID="<skill-name>"` parses in both, and the quoted placeholder then reaches the `REFUSED` branch this PR added — strictly better than a concrete example.

**F3 — three of the four divergence-table rows were wrong,** in a table whose stated contract is that each row names the direction it errs. Re-derived with the shipped stripper lifted out of `SKILL.md` and the pre-#734 loop lifted out of this PR's diff:

| fixture | raw | CommonMark | this block | pre-#734 |
|---|---:|---:|---:|---:|
| `~~~text` wrapping a comment | 9 | 9 | **6** | 6 |
| unclosed ```` ```text ```` | 5 | 5 | 5 | 5 |
| unclosed fence then a comment | 8 | 8 | 8 | 8 |
| `<!--` with no `-->`, containing ```` ``` ```` | 6 | 1 | **1** | **5** |

- `~~~` was listed **unmeasured**; it under-reports, the direction this document itself calls dangerous
- "an unclosed fence" was listed **not handled**; it is handled and correct
- the unclosed **comment** was absent, and this repair changed it from 5 kept to 1 — the only unbounded strip path left, which finding 1 widened the inputs to

**My first run of this got it wrong**, and the correction is worth stating: I reconstructed the pre-#734 loop from memory of its shape and it reported no behaviour change on the fourth fixture. The real loop folds `cmt` into the `elif` under `not fence`, which is exactly the interaction at issue. The instrument is now taken from the diff hunk, not retyped — the reviewer had it right and I briefly had it wrong.

The rationale paragraph is rewritten too. "Repairing toward CommonMark cannot make things worse" is **false** — a `~~~` fence containing an unterminated comment also strips to EOF. The sound distinction, which the reviewer supplied, is that finding 1 stripped ordinary index lines outside any comment and any fence, content no plausible loader grammar drops; every remaining divergence strips a region beginning on a `<!--` line.

**F5 — the byte-identity gate could be disarmed by following its own error message.** The reviewer confirmed the empty-vs-empty pass I asked about is genuinely closed. The residual hole is different: `extractBlock`'s failure text tells the reader to drop a file from `CARRIERS` when the block is gone on purpose. Doing that for `budget`'s second carrier leaves one entry, the loop skips `SOURCE_OF_TRUTH`, the body runs zero times, and the test passes having compared nothing. Now asserted.

### Mutant, and one that survived

| mutation | result |
|---|---|
| `budget: [SOURCE_OF_TRUTH, 'skills/manage-memory/SKILL.md']` → `[SOURCE_OF_TRUTH]` | **KILLED by 1** |

A **first attempt survived**, and it indicted my needle rather than the code: `'skills/manage-memory/SKILL.md',` with a trailing comma matches the `reachability` list, not `budget`. Removing one of `reachability`'s four still leaves three, so the new assertion correctly did not fire. Stated plainly because it is a real limit of the fix: `length > 1` closes the **vacuous** case, not the incomplete one. Dropping one of `reachability`'s four carriers is still uncaught, and pinning the lists exactly would be the stronger gate.

### Verification at `d5f08a487`

- `npm run test:scripts` — **863/863**, 0 fail
- `npm run validate:integrity` — **PASSED** (7 warnings, pre-existing glyph/palette)
- `npm run check-readmes` — all files up to date
- No mirror work: the four mirrors carry `SKILL.md` only, and `SKILL.md` is unchanged by this commit

Evidence: `tests/results/2026-09-16-divergence-table-recheck/`

### Not fixed, deliberately

**F4 — 494 is a derived constant with no owner.** The arithmetic is right (English 491, mirrors 497, delta 6), but nothing computes or asserts it; a seventh provenance field silently makes it 493 with nothing going red, and `refresh:stubs` drops `fence_basis_commit` until `--stamp` runs, so the overhead is 5 in that window. The reviewer also notes the two-tree scan does not protect a **new** skill, whose mirrors are scaffolded after merge and so do not exist when the local pre-check runs. The durable fix derives the ceiling per English file rather than writing a number in a paragraph. Filing rather than bolting on.

**F8 cosmetics** — no in-file marker naming `memory-blocks.test.js` at the duplicated fences, and the mirrors' `source_commit` moved from a short sha to a quoted 40-character one, which is tool behaviour but now inconsistent with the rest of the corpus.

### Weakest claim, named

The whole divergence analysis is scored against **CommonMark**, and the target is the loader, whose grammar nobody has captured. Where the two disagree the table records a direction, not a defect — and for the unclosed comment the new behaviour is right under CommonMark and *worse* under a loader implemented as a non-greedy `<!--.*?-->`. `tools/wirecap.py` plus the generator in `tests/results/2026-08-25-fence-strip-replication/` is the arm that would settle it, and it has not been run.

Issue #845, opened 2026-09-15
[MED] i18n: write-continue-here's ten mirrors have been outside fence checking since 1.1 — a js fence they never received


pjt222:
## Summary

All ten `write-continue-here` mirrors have been outside frozen-fence checking since the skill went
to 1.1, and nothing said so. English gained a ` ```js ` fence — the `verify-handoff` Workflow call
in Step 3 — which no mirror received. The folded tag sequence therefore differs, ordinal mapping is
unsound, and `check:fence-propagation` declines to compare rather than reporting a divergence:

```
$ npm run check:fence-propagation -- --id write-continue-here
write-continue-here (skills): 0 of 10 mirror(s) alignable, 2 frozen of 3 fence(s)
  UNALIGNABLE i18n/<locale>/skills/write-continue-here/SKILL.md — folded tag sequence differs
  from English — ordinal mapping is not sound, not compared
  10 finding(s)
```

Measured at `ee55ed509`, before any of #468's work. This is not a regression introduced by that PR;
it is the state that PR found.

## Why it is invisible

Three gates could each have caught it and none does, for three different reasons:

- `check:fence-propagation` is **not in CI** and is id-scoped with no default, so it only runs when
  someone names this skill.
- `validate:i18n-fences` runs `--warn`, and its tag-structure findings are ratcheted by class. This
  one lands in the **unjudged `unalignable` pool**, which produces no finding at all (#628 describes
  the same blind spot for a different cause).
- `check-translation-freshness` already reports these mirrors `STALE`, and has for months. Staleness
  is the normal, expected state of a mirror after an English edit, so the signal that would
  distinguish "prose is behind" from "a frozen fence is missing entirely" is drowned by the one that
  is true of most of the corpus.

## The numbers

| | English | every mirror |
|---|---|---|
| lines | 186 | 137–140 |
| fences | `bash`, `bash`, `markdown`, `js` | `bash`, `markdown` |
| `source_commit` | — | `025eea68` (10/10) |
| `fence_basis_commit` | — | `025eea68` (10/10) |

The mirrors are now two English revisions behind: 1.1's `js` fence, and 1.2's shared resolver
(`bash`) from #468.

## Why this was not fixed in the #468 PR

Restoring alignment means **inserting** two fences into ten translated files, each needing a lead-in
sentence explaining what the fence is. Writing those sentences in German, Japanese, Chinese and the
caveman/wenyan registers is a translation job, not a byte propagation, and lifting the English
[...]

Issue #775, opened 2026-09-02
write-continue-here enforces gitignoring the handoff, which breaks the tracked-handoff lifecycle and its own delete-after-use rule


pjt222:
## Summary

`skills/write-continue-here` treats gitignoring the handoff as a **validation requirement**, not a choice. That is wrong for any project where the handoff has to travel — between machines, to a cloud session, or into review — and it actively fights a lifecycle that some projects have deliberately adopted.

Two places state it:

- Step 2, On failure: *"Verify `.gitignore` contains `CONTINUE_HERE*.md` — the pattern must cover the draft as well as the installed file — and if not, add it."*
- Validation checklist: *"`.gitignore` includes `CONTINUE_HERE*.md` (the draft and the installed file)"*

So a project that tracks its handoff **fails the skill's own validation**, and a session following Step 2's failure branch will silently add an ignore rule that breaks the project's workflow.

## The counter-case, from a project that runs this lifecycle

`claude-buddy-private/CLAUDE.md` § "Handoff lifecycle" states the opposite as a deliberate rule:

> It stays **tracked, not gitignored**. Tracking is what makes deletion safe — `git log -p -- CONTINUE_HERE.md` recovers every version ever written — and it lets cloud sessions and other machines receive the handoff at all.

That project removed a pre-existing `.gitignore` entry for exactly this reason, having found it had been silently failing every `git add --force`. Both properties are real:

- **Recoverability makes the ephemeral lifecycle safe.** The skill tells you to delete the handoff after consuming it. Deleting an ignored file destroys it; deleting a tracked file leaves it in history. The skill's own delete-after-use rule is *safer* under tracking, so the two requirements are in tension inside one skill.
- **A background, cloud, or second-machine session cannot receive an ignored file at all.** It never leaves the originating checkout.

## What is right about the current default

Untracked is the better default for a solo local project: the handoff is session scratch, and committing it adds noise to the log. This issue is not asking to invert the default — it is asking to stop enforcing it.

## Suggested fix

1. Demote both mentions from requirement to **documented choice**, with the trade-off stated: untracked keeps the log clean; tracked makes deletion recoverable and lets other machines and cloud sessions receive it.
2. Remove the "if not, add it" instruction from Step 2's failure branch. A skill should not mutate a project's `.gitignore` as a side effect of writing a handoff.
[...]

Issue #631, opened 2026-08-17
i18n: a frozen fence edited in English silently orphans its ten mirrors — no gate can see it


pjt222:
Found while doing #551, by the instrument built to verify that fix (`scripts/check-fence-propagation.js`, added in the same PR). Measured, not inferred.

## The hole

`check-i18n-fence-parity.js` accepts a gated fence body that appears in **any** revision of the paired English file. That is deliberate and correct — 2,571 translations are stale and `evolve-*` bumps `source_commit` without retranslating (#405/#616), so anchoring to HEAD would make the gate unusable.

The consequence is exact:

> Edit a frozen fence in English, propagate to **zero** mirrors, and the parity gate reports `violations: 0` — before and after the commit. The old body is in history, so it is in the pool, forever.

Measured on `skills/write-helm-chart`: after the English-only edit, `node scripts/check-i18n-fence-parity.js --id write-helm-chart --json` returned `violations: 0, ungatedDivergences: 0`.

`check-translation-freshness.js` carries no information here either. Those ten mirrors were **already** `STALE` before the edit (`translated at 33b561c9, source now at f818c698a`), so the edit moved no signal at all. Staleness fires for 2,571 files and therefore says nothing about whether *this* fence was propagated.

Both gates are green whether you propagate to ten mirrors or to none.

This is [`feedback_historical_match_gates_miss_deletions`](https://github.com/pjt222/agent-almanac/issues/477) reached from the **edit** side rather than the deletion side: a historical-match rule can never see the source move on. It is adjacent to #628 (`unalignable` is an unjudged pool) — both are places where the gate's silence is not evidence of health.

## Why it matters more than ordinary staleness

Prose staleness is recoverable: a translator reads English and retranslates. A frozen fence is **not** translatable by definition — the mirror must carry English bytes verbatim. So an unpropagated fence edit is not a translation debt anyone can pay by translating. It is a byte copy that simply never happened, and nothing asks for it.

CLAUDE.md already records the inverse hazard — a line added inside a frozen fence becomes ten permanently-untranslatable copies. This is the same fence, the other direction: a line added in English becomes ten mirrors that permanently disagree, invisibly.

## It is not a one-off

[...]

Issue #502, opened 2026-08-07
i18n: 196 comment-only frozen fences carry instructions the #472 rule strips from every translation


pjt222:
## The gap

#472 freezes non-`text`/`markdown` fences to English, comments included, and states the remedy: *"If a comment carries the only statement of an instruction, lift it into the step's prose rather than translating it in place."*

The freezing is now mechanised across four batches (#495, #496, #499, and the `python` batch in progress). **The lifting has never been done, and nothing tracks it.**

Measured across `i18n/`, gated fences of ≥3 non-blank lines that are ≥90% comment lines:

| | count |
|---|---:|
| comment-only gated fences | **196** |
| files affected | **106** |
| by tag | bash 104 · r 60 · python 20 · json 10 · dockerfile 1 · yaml 1 |

Not all 196 are load-bearing — plenty are incidental (`# Load data`). The ones that matter are where the comment is the only statement of the step, and separating those needs human judgment.

## Worst-case exemplar

`i18n/<locale>/skills/rotate-scraping-proxies/SKILL.md`, Step 1 "Pre-flight Legality and Ethics Check", in de/es/ja/zh-CN. The fence is 100% comments and carries the entire substance of the step:

```python
# Inputs to confirm before writing any code:
# 1. Is the data public (no login required)?
# 2. Does robots.txt permit the path?
# 3. Does the site's ToS prohibit automated access? (read it)
# 4. Would the scraping process personal data? If yes, what is the legal basis?
# 5. Could this access circumvent geo-licensing, paywalls, or auth?
# 6. Is there a public API or data dump that would make scraping unnecessary?
# 7. Have you contacted the site owner if scope is large?
```

The German prose around it says only *"Machen Sie den gesamten Arbeitsablauf von einer dokumentierten rechtlichen und ethischen Prüfung abhängig"*, and the **Expected:** block reads *"Jede Frage hat eine verteidigungsfähige schriftliche Antwort"* — referring to seven questions that no longer exist in German anywhere in the file. Before the restore, they did.

This is GDPR legal basis, ToS compliance and geo-licensing circumvention, in the one step the skill itself calls "die mit Abstand größte Schadensquelle". English readers are fine — English carries it in the fence. Non-English readers lost it.

## Why this is not a reason to revert

[...]

PR #846, opened 2026-09-15
fix(skills): resolve the handoff instead of hardcoding the root, and repair a SessionStart hook that never injected anything

### LINKED ISSUE #844: [HIGH] read-continue-here: the shipped SessionStart hook emits a JSON shape Claude Code discards, so it has never injected anything
## Summary

`read-continue-here` Step 6 installs a SessionStart hook whose final line emits

```json
{"hookSpecificOutput":{"sessionStartContext":{"additionalContext":"…"}}}
```

`sessionStartContext` is not a key in Claude Code's hook output contract. The hook finds the
handoff, strips CRLF, JSON-escapes the content, prints it — and Claude Code discards the whole
object. The hook has never injected anything, at any path, in any project.

This is the same failure class as #468 and one layer beneath it: #468 reports that a handoff kept
somewhere other than the project root is silently ignored. Measured, a handoff kept **at** the
project root is silently ignored too.

## Findings

1. **[HIGH]** The published heredoc emits an unknown key (`skills/read-continue-here/SKILL.md:183`).
   The documented shape puts `additionalContext` **directly under** `hookSpecificOutput`:

   > Return `additionalContext` inside `hookSpecificOutput` alongside the event name

   and the per-event table lists `SessionStart` as `hookSpecificOutput.additionalContext`
   (Claude Code hooks reference, fetched 2026-09-15). `sessionStartContext` occurs **zero** times
   in that document.

2. **[HIGH]** A malformed object is not loud. Per the same reference, stdout that starts with `{`
   and fails schema validation is reported as a *non-blocking* error and the content is not added
   as context. So the failure surfaces to the user as "the agent didn't seem to know what I was
   working on" — exactly the symptom #468 describes for the wrong-path case.

3. **[MED]** The repository's own `.gitignore:22–26` and `write-continue-here` assume this pipeline
   works end to end. Any consumer who followed Step 6 has a hook that runs on every session start
   and does nothing.

## Measurement

Three arms, one random sentinel per arm generated at run time and never named in the prompt, so a
model reporting it must have read it rather than reconstructed it. Claude Code 2.1.272, Linux/WSL2.
Reproduced twice with fresh tokens.

| arm | hook stdout | model saw the sentinel |
|---|---|---|
| C (control) | plain text `SENTINEL=…` | yes |
| A (documented) | `{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"SENTINEL=…"}}` | yes |
| B (**as shipped**) | `{"hookSpecificOutput":{"sessionStartContext":{"additionalContext":"SENTINEL=…"}}}` | **no — answered NONE** |

Arm C is the control: it proves the harness can carry a sentinel from a SessionStart hook into the
model at all, so arm B's negative is a property of the shape, not of the probe. Each arm builds its
own `mktemp -d` git repo and passes its own settings file via `claude -p --settings`; nothing in the
probe touches this repository or `~/.claude`.

Probe script and both captured runs land under `tests/results/2026-09-15-sessionstart-hook-shape/`.

## Fix

One line of the heredoc:

```bash
{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":$ESCAPED}}
```

`hookEventName` is not required by the reference but every working hook in the wild carries it and
the documented example does; include it.

The fix is cheap. What is not cheap is that nothing in this repository could have caught it — the
heredoc is a frozen fence whose *bytes* are checked across ten mirrors while its *contract* is
checked nowhere. A test that runs the emitted JSON against the documented shape is the durable half.

## Acceptance criteria

- [ ] The heredoc emits `additionalContext` directly under `hookSpecificOutput`
- [ ] A test extracts the hook body from `SKILL.md`, runs it against a fixture handoff, and asserts
      the emitted JSON matches the documented shape — failing on the shipped shape as the negative arm
- [ ] The corrected bytes reach all 10 `i18n/*/skills/read-continue-here/SKILL.md` mirrors in the
      same commit, verified by `npm run check:fence-propagation -- --id read-continue-here`
- [ ] The probe and its captured runs are committed as evidence, with the Claude Code version stated
- [ ] Consumers who installed the old hook are told to re-run Step 6 — a note in the skill, since the
      installed copy lives outside this repository and cannot be migrated from here

## Context

Found 2026-09-15 while rewriting Step 6's heredoc for #468 (path resolution). #468's AC2 — the hook
must distinguish "no handoff exists" from "a handoff exists somewhere I did not look", and say so —
is unsatisfiable while the channel that would carry the message is discarded, so the two are fixed
together.

Related: #468, #775, #660.

pjt222:
Closes #844. Addresses #468 ACs 1/2/3/6 and #775 AC4's line. #468's ACs 4/5 and the rest of #775 are W3-P2's, deliberately.

## What this changes

`read-continue-here` 1.0 → 2.0, `write-continue-here` 1.1 → 1.2.

Both skills resolve the handoff through **one shared block** — `CONTINUE_HERE.md`, `docs/CONTINUE_HERE.md`, `.claude/CONTINUE_HERE.md`, in that order, anchored on the repository root rather than `$PWD`. The block is byte-identical in both files, which is a claim about two hand-maintained copies of shell in two markdown files that nothing in this repository could previously have noticed drifting: the i18n fence gate compares a mirror against its English source, never two English files against each other. `scripts/test/continue-here-blocks.test.js` is that owner.

`read-continue-here` Step 5 now deletes the file it actually read, branching on whether the project tracks it — `git rm` plus a commit when tracked, plain `rm` otherwise. Both lifecycles are legitimate; #775's point is that the skill must not enforce one.

Step 6's SessionStart hook is rewritten, and that is the part worth reading.

## The hook has never injected anything

1.x emitted `additionalContext` nested inside a `sessionStartContext` object. That key does not exist in Claude Code's hook contract — **zero occurrences** in the hooks reference, fetched and grepped rather than summarised, which gives the shape as `hookSpecificOutput.additionalContext`. An object that starts with `{` and then fails schema validation is reported as a *non-blocking* error: the session continues and the context is dropped. So the hook read the file, stripped CRLF, JSON-escaped it, printed it, and Claude Code threw it away, in every project, at every path, for the life of the skill.

#468 read the wrong-path case as the bug. The wrong-path case was one symptom of a channel that was never open, and its AC2 — *distinguish "no handoff exists" from "a handoff exists somewhere I did not look", and say so* — could not have been satisfied while the thing carrying the message was discarded. That is why #844 is fixed here rather than separately.

## Evidence

`tests/results/2026-09-15-sessionstart-hook-shape/`. Three probes, all running the **published bytes** — extracted from the SKILL.md by its heredoc delimiters, with the 1.0 arms extracted from `ee55ed509`, the last revision carrying 1.0, the same way — never a retyping. Every sentinel is random per run and never named in the prompt, so an answer carrying it was read, not reconstructed.

| probe | what it settles |
|---|---|
| `hook-shape-probe.sh` | control green, documented shape green, **shipped shape NONE**. Reproduced twice with fresh tokens. The control is what makes the negative mean something. |
| `hook-arms.sh` | the 2.0 heredoc across eight layouts — each supported path, an unsupported one, none, a subdirectory cwd, content needing JSON escaping, and a machine with no `jq`. Plus a negative arm: the same validator rejects the 1.x object, naming the key it found. |
| `ac6-docs-layout.sh` | **#468 AC6** end to end. Five real headless sessions. `docs/` layout with 2.0 injects; with 1.0 it does not; root layout with 1.0 does not either; a handoff at `notes/` is reported by 2.0 and the model recovers it; the same layout under 1.0 answers `NONE`, which is the control proving the report is load-bearing rather than something the model would have found unaided. |

The fourth AC6 arm's first matcher was wrong and is recorded in `RESULT.md` rather than quietly fixed: it looked for the report's wording in the answer and found the sentinel instead, because the model had been told where the stray handoff was and opened it. The outcome AC2 asks for, scored as a failure by a matcher checking the wrong thing.

## Proving the new gate can fail

Four mutants against `npm run mutation-check`, each restored:

| mutation | result |
|---|---|
| hook JSON back to the 1.x `sessionStartContext` nesting | **killed by 3** (`--allow-broad`: three independent tests read the payload) |
| one candidate path changed in `write-continue-here`'s copy of the resolver | **killed by 1** — the drift gate |
| Step 5's fail-closed guard given a hardcoded `CONTINUE_HERE.md` fallback | **killed by 1** — it deletes a file it never resolved |
| `cleanEnv` removed from the suite's own `git` calls | **killed by 1** — but only after the arm was repaired; see below |

That last one **survived** on first run, and the survivor indicted the fixture rather than the code.
The isolation was correct; the arm could not see it, because the victim repository and the fixture
used the same git identity, so a leaked `git config` rewrote identical bytes and the tree digest did
not move. Giving the victim a distinct identity made the arm able to fail, and it now does.

## Premises

Every row marked False is a premise I held or the plan stated, and acted on.

| premise | source | holds? | how it was checked |
|---|---|---|---|
| the 20 mirrors are a mix of stubs and translations | the plan's W3-P1 row | **False** | `refresh:stubs --verify` refuses 10 of 10 on each skill. Zero stubs; all 20 are hand work with no tool assist. |
| the hook works and only its path is wrong | #468's framing | **False** | measured: never injected anything, at any path (above) |
| `git ls-files --error-unmatch` separates tracked from ignored | assumed while writing Step 5 | **False** | git 2.43: `0` tracked, `1` untracked, `1` **ignored**, `1` missing, `128` outside a repo. Harmless — four states, one correct arm — but the prose says the measured thing, not the assumed one. |
| a fence that runs in this session runs as published | — | **False** | `: "${VAR:?… Step 1's resolver …}"` is a **bash syntax error** (`unexpected EOF while looking for matching '`) and valid zsh. This session's shell is zsh, so it could only surface when the test ran the fence under bash. All three guards reworded. |
| adding a verification fence to Step 6 is free | my first draft | **False** | it changes the folded tag sequence, which would turn 10 currently-alignable mirrors UNALIGNABLE and lose gate coverage. Converted to prose; the fence count is unchanged. |
| `source_commit` must not move on an English-only edit | `evolve-skill` Step 4.5 (`:158`, `:182`) | **True** | read in the skill, then verified in the diff: no `source_commit` line changed in any of the ten. |
| `normalize-i18n-fences.js` could do the propagation | — | **False** | it restores from `source_commit`, the revision *before* the edit. Recorded on #631 with what the missing tool would do. |

## Scope

- **`write-continue-here`'s ten mirrors are left exactly as found.** They have been UNALIGNABLE since 1.1 — English gained a ` ```js ` fence they never received — measured at `ee55ed509` before any edit here. Restoring them means inserting two fences into ten translated files with lead-in prose in ten registers; lifting the English sentences in instead is the defect class #502 exists for. Filed as **#845**. This means the wave's exit line "read/write-continue-here fences identical across 22 files" is met for 11 of the 22 and cannot honestly be met for the rest in this PR.
- **`write-continue-here:99` is split between PRs.** Its "project root" clause is fixed here, with the rest of #468's root class; its `.gitignore` enforcement sentence is #775's and stays for W3-P2, which is where the tracked-versus-ignored trade-off becomes a numbered step.
- The installed CLAUDE.md snippet at Step 7 is edited. It sits in a ` ```markdown ` fence, so it is exempt from byte-propagation but not from being correct.
- **The suite's own git calls are isolated from the caller's environment.** `cwd` and `-C` are not isolation — git honours an absolute `GIT_DIR` over both, and this suite runs `git init`, `git config`, `git rm` and `git commit`. Every `GIT_*` key is dropped and `HOME`/`XDG_CONFIG_HOME` move to the fixture, with an arm that points `GIT_DIR` at a victim repository and hashes its whole tree, `.git` included, before and after.
- **`ac6-docs-layout.sh` pins its 1.0 baseline to `ee55ed509`,** not `origin/main`. After this merges `origin/main` carries 2.0, and a probe reading it would extract the *fixed* hook into the three arms that expect `NONE` — passing while asserting the opposite of what it says.

## Checks

All re-derived at `84c331eec`, the head this body describes. `validate:integrity` (7 warnings, all pre-existing viz glyph/palette rows), `validate:line-endings`, `check-readmes`, `check:generated-artifacts`, `ratchet` (no new member), `validate:content-style --added` (no offender in these files), `test:scripts` **874/874**, `check:fence-propagation --id read-continue-here` → `OK — 6 frozen fence(s) byte-identical across 10 mirror(s)`, from 50 `LAGS` findings. CI: **16 pass, 0 fail**.

`evolve-skill` Step 4.5 point 1, run rather than assumed: `validate:translations` reports **all 20** mirrors of the two skills `STALE` against `c248c5f0c`, their own `source_commit` unmoved at `025eea68` and `82c77053`. That is the signal being left deliberately true.

`skill-path-refs.test.js` went red on the new cross-reference and its pinned set was updated rather than loosened — the gate doing its job.

## The weakest claim here

**That the rewritten hook is right for everyone, not just for this machine.** The shape is measured on one Claude Code version (2.1.272) and one platform (WSL2). Nothing establishes that `sessionStartContext` was never valid — only that it is not valid now. And the stray-handoff report's *usefulness* rests on one model's behaviour on one prompt; that it is emitted is mechanical, that it is acted on is a single observation.

Second weakest: `find -maxdepth 3` bounds the stray search because a SessionStart hook runs under a timeout. A handoff at depth 4 is still reported as absent — the old failure, moved deeper rather than removed.

No agent review round: the plan's W3-P1 row carries no `[REVIEW]` marker, Fable credits are exhausted, and this is not `docs-only` — it touches two skills, ten mirrors, two test files and an installed hook.


pjt222:
Handover. Green 16/16 at `84c331eec`, no agent review round — the plan's W3-P1 row carries no `[REVIEW]` marker, Fable credits are exhausted, and this is not `docs-only`: two skills, ten mirrors, two test files, and a hook that runs on every consumer's session start.

**The weakest claim, named rather than buried:** that the rewritten hook is right for everyone, not just for this machine. The shape is measured on one Claude Code version (2.1.272) and one platform (WSL2). Nothing here establishes that `sessionStartContext` was *never* valid — only that it is not valid now, and that the documented key is. If you know it once worked, that changes the framing of #844 from "always broken" to "broken by a contract change", and the skill should say so instead.

Second weakest: the stray-handoff report's *usefulness* rests on one model's behaviour on one prompt. That it is emitted is mechanical and covered by a test; that it is acted on is a single observation, with a control.

**Two things that need you rather than me:**

1. Your installed `~/.claude/hooks/continue-here/read-continuation.sh` carries the same dead shape, and this repository has a 145 KB `CONTINUE_HERE.md` at its root, modified today, that no session has ever been handed. The fix is one line; I have not touched it, because the installed copy is the consumer's and not this PR's.
2. The wave's exit line — *read/write-continue-here fences identical across 22 files* — is met for 11 of 22 and cannot honestly be met for the rest here (#845 explains why). Either accept 11/22 for this wave, or fold #845 into W3-P2 where the same files are already open.

pjt222:
## Adversarial round ran (Opus, Fable being out); two blocking findings fixed at `7aaddc512`

The round happened after all — Fable credits are exhausted until Sunday, and the operator's ruling was to re-route the reviewer to Opus rather than skip it. `advocatus-diaboli` reviewed the bundle cut at `84c331eec` (`BUNDLE_STATUS` clean) and returned three blocking findings. It steelmanned the PR first and killed 9 of 11 of its own mutants against the suite, so the two it found vacuous are worth taking seriously.

**Both blockers re-derived independently before I touched anything** — I did not take the reviewer's word for either.

### B1 — the `awk` fallback emitted invalid JSON for every C0 byte but tab (fixed)

`jq -Rsa .` escapes the whole control range; the fallback handled backslash, double quote and tab and let the rest through raw. JSON forbids every raw byte below 0x20 inside a string, so the object stopped parsing — and an unparseable object is discarded in exactly the silent, non-blocking way #844 was about. 1.x had the channel closed so this could not surface; 2.0 opens it, and `jq` ships in neither a stock Ubuntu nor a macOS install, so `awk` is the common path.

Measured against `jq` across fifteen inputs on **both** GNU Awk 5.2.1 and mawk 1.3.4 (mawk matters — it is Debian/Ubuntu's default):

| escaper | gawk | mawk |
|---|---|---|
| as shipped | **9/15** | **9/15** |
| repaired | **15/15** | **15/15** |

The six failures are a mid-line CR, ESC, SOH, BS/VT/FF and an all-controls line. Both realistic here: a handoff quoting terminal output carries ESC (`FORCE_COLOR=3` is set on this machine) and one written on NTFS carries CR.

Probe, negative-arm control and full transcript: `tests/results/2026-09-16-hook-escaper-c0/`. It also records an instrument correction that inverted the first reading — the probe initially scored **0/15 including a plain-ASCII input**, which was the tell that the ruler was wrong rather than the subject: `jq` keeps a file's trailing newline and the awk program drops it, while the real call path strips it in command substitution before either sees it.

### B2 — Step 5's commit carried no pathspec, so it committed the whole index (fixed)

`git commit -m` with no pathspec commits everything staged. Measured in a fixture with two unrelated files staged:

```
msg=chore: consume the session handoff
 CONTINUE_HERE.md | 1 -
 other/thing.txt  | 1 +
 src.txt          | 1 +
```

Exit 0, silent. This runs at **session start** — precisely when a prior session or a peer sharing the worktree may have left work staged — and this repository's own `CLAUDE.md` mandates pathspec-scoped commits for that reason. Fixed with an explicit pathspec; the commit now names one path and the unrelated change stays staged.

### B3 — `write-continue-here`'s ten mirrors: deferred, to #845

The reviewer's own option, taken deliberately. Those mirrors have been UNALIGNABLE since 1.1 over a `js` fence they never received; this PR's resolver makes it two fences rather than one. Re-measured on the current head: **`2 frozen of 3` at `ee55ed509` became `3 frozen of 4` now**, 0 of 10 alignable, 10 findings.

Not folded in, because closing it is a **translation** job — two fences into ten files across six registers, each needing a lead-in sentence, and lifting the English sentences in is the defect class #502 exists for. It is also sequenced wrong: W3-P2 edits this skill's English next, which would re-stale all ten immediately. Recorded on #845 with the reasoning, blocked on W3-P2. The wave exit line is accepted at 11 of 22.

### The two vacuous arms, also closed

The same round found two of the suite's own assertions passing for the wrong reason — worth more than either blocker, since they were written to prove things they did not prove:

- the unset-path arm asserted a non-zero exit, which `rm ""` supplies on its own once the guard is deleted. It now asserts the guard's own message.
- `RESOLVER_CARRIERS` was not pinned, so reducing it to a single entry made the byte-identity comparison compare a file to nothing and pass. Now pinned at two.

### Every new assertion proven able to fail

`npm run mutation-check`, four mutants, each **killed by exactly one test** — the count matters, since a large one usually means a crash rather than coverage:

| mutation | result |
|---|---|
| disable the control-character loop in the awk escaper | KILLED by 1 |
| drop the pathspec from Step 5's commit | KILLED by 1 |
| delete Step 5's unset-path guard | KILLED by 1 |
| reduce `RESOLVER_CARRIERS` to one entry | KILLED by 1 |

The guard mutant's needle carries the following line, because that guard appears three times and a one-line needle hit all of them — `mutation-check` refused it rather than crediting a collateral kill, which is the refusal working.

### Verification at `7aaddc512`

- `npm run test:scripts` — **874/874**, 0 fail
- `npm run validate:integrity` — **PASSED** (7 warnings, all pre-existing glyph/palette, untouched here)
- `npm run check:fence-propagation -- --id read-continue-here` — **OK, 6 frozen fences byte-identical across 10 mirrors**
- Propagated with `tools/patch-literal.py`, both needles verified to occur exactly once in all eleven files before anything was written; `fence_basis_commit` restamped to `a2c36f4e6`, `source_commit` verified unmoved by diff

### Non-blocking, not fixed here

Left deliberately, all reported by the same round: BSD/macOS `sed` treats the CR strip as a literal `r` so it would trim ordinary words (predates 2.0, no Mac available to confirm); the earlier probe's arms A and B differ in two ways, so they cannot attribute the 1.x failure to the nesting *specifically*, only establish that it failed; Step 5 cites a trade-off `write-continue-here` does not yet state, which is W3-P2's job; and the 2.0 upgrade note mentions the hook but not that Step 5 may now write history.

### Weakest claim, named

That the hook fix is right for *everyone* still rests on one Claude Code version (2.1.272) and one platform, and nothing establishes the 1.x shape was never valid — only that it is not valid now. B1 narrows the exposure rather than removing that caveat: valid JSON is correct under either explanation, but the *severity* argument for B1 rests on Claude Code discarding malformed stdout, which is read from the hook reference and from #844's observed behaviour, not from a run that feeds it malformed JSON on purpose. The settling experiment is named in `RESULT.md`.

Issue #853, opened 2026-09-16
Decide: write tools/enforce-redaction-gate.sh, or drop the gate step from the three skills that name it


pjt222:
## Summary

`tools/enforce-redaction-gate.sh` is named by three separate skills (`enforce-redaction-gate`, `redact-wire-capture`, `redact-visualization-for-disclosure`) as the verification step each transform ends on, and has never existed in this repository — confirmed again while working W3-P3 (#751). PR fix/751-redaction-tooling reframes the fence invocations as templates the reader must build rather than shipped commands, and states this plainly in prose, but does not resolve the underlying gap: three skills structurally depend on a gate that does not exist.

## Findings

1. **[MED]** `enforce-redaction-gate/SKILL.md` itself describes, in its own Steps 1-4, exactly what the script would need to be (a two-tier shape + structure-aware scanner under a strict exit-code contract) — the specification exists, the implementation does not.
2. **[MED]** `redact-wire-capture` Step 4 and `redact-visualization-for-disclosure` Step 5 both end their procedure by handing off to this nonexistent script for the structure-aware verification tier; without it, both skills' own Validation checklists (`enforce-redaction-gate exits 0 on the scrubbed directory`) are unverifiable as written.
3. **[LOW]** `tools/check-redaction.sh` (the shape-tier scanner `redact-for-public-disclosure` names) DOES ship and has a real, tested `--verify` mode — so the four-skill family is inconsistent: one skill's named tool is real, three skills' named tool is aspirational.

## Acceptance Criteria

- [ ] A decision is recorded on this issue: write `tools/enforce-redaction-gate.sh`, or drop the gate step from the three skills that reference it
- [ ] **If write:** `tools/enforce-redaction-gate.sh` ships with a `--verify` self-test per this repository's `tools/` convention (see `tools/README.md` § Adding a Tool), gets a row in `tools/_registry.yml`, and the four `scripts/check-skill-path-refs.js` ALLOWLIST entries naming it (added in #751/W3-P3) are removed as stale
- [ ] **If drop:** the three skills' verification steps are rewritten to state their own inline verification instead of handing off to `enforce-redaction-gate`, the corresponding ALLOWLIST entries and the now-orphaned fence bodies are removed together, and `enforce-redaction-gate`'s own description — currently "the verification step every redaction transform ends on" (echoed in its Related Skills section and in `redact-wire-capture`'s and `redact-visualization-for-disclosure`'s Related Skills sections) — is corrected across all four skills

[...]

Issue #852, opened 2026-09-16
Decide: write tools/redact-visualization.py, or drop the invocation from redact-visualization-for-disclosure


pjt222:
> **Rewritten 2026-09-16.** The original body was built on a premise its filer was given and that was false — see [the correction comment](https://github.com/pjt222/agent-almanac/issues/852#issuecomment-5697387213). Three of the four skills have **zero** i18n mirrors, so the fence bodies were never frozen and #631 was never a blocker. PR #854 then corrected them in place. What survives is narrower and is stated below. The filer's original findings list is preserved verbatim at the bottom.

## Summary

Four redaction skills named tooling that has never existed (#751). PR #854 (`e1f1413b8`) corrected the presentation **inside** the fences in the three mirror-free skills, so no fence still reads as a bare runnable invocation of an absent script.

What remains is the underlying fact rather than its presentation: **`tools/redact-visualization.py` does not exist**, and `redact-visualization-for-disclosure` Step 3 is the only call site left whose named script has no decision attached to it. `tools/enforce-redaction-gate.sh` — the other absent script, named by three skills — is #853's subject, not this issue's.

## What is already done

| | |
|---|---|
| `enforce-redaction-gate` Step 4 + Step 5 fences | corrected in place (#854) |
| `redact-wire-capture` Step 4 fence | corrected in place (#854) |
| `redact-visualization-for-disclosure` Step 3 + Step 5 fences | corrected in place (#854) |
| `redact-for-public-disclosure` | never in scope — its 7 fences are byte-identical to `origin/main`, its 10 mirrors untouched, and `tools/check-redaction.sh` ships |

Each corrected fence now carries a comment naming what does not ship and pointing at the issue that decides it, and each Expected block carries a "not yet checkable here" clause. Verified in round 2 of #854 by comparing every fence in all four files against `origin/main` by tag, body SHA-256, count and closed-ness.

## What remains

1. **`tools/redact-visualization.py` has no write-or-drop decision.** `redact-visualization-for-disclosure` Step 3 invokes it; Step 2 already specifies the ordered longest-first mapping the script would implement. Either write it — Step 2 is a complete enough spec that the script is mostly transcription — or drop the invocation and leave Step 2's mapping as the procedure the reader applies by hand. #853 made exactly this decision for `enforce-redaction-gate.sh`; nobody has made it for the `.py`.

[...]

Issue #801, opened 2026-09-07
i18n: 1,041 wenyan-tier files carry the caveman homage as their translator — the attribution names the wrong tier


pjt222:
## Summary

Every translated file under `i18n/wenyan`, `i18n/wenyan-lite` and `i18n/wenyan-ultra` carries `translator: "Julius Brussee homage — caveman"` — the caveman tiers' attribution string, on files whose bodies are classical Chinese. The field is a signal for humans ("who translated this?") and it names the wrong tier for a whole locale family. Found while reviewing #800, where the refresh tool's refusal message quoted the field on wenyan mirrors of `validate-references` and the reviewer asked whether those files were translated at all (they are: the wenyan body begins `# 驗參考` / `察 BibTeX 之目以求全、準、一。…`, the caveman body `# Validate References` / `Check BibTeX bibliography entries for completeness, accuracy, consistency.`).

## Findings

1. **[MED]** `grep -rl 'Julius Brussee homage — caveman' i18n/wenyan i18n/wenyan-lite i18n/wenyan-ultra | wc -l` → 1041 files; only 5 files under `i18n/wenyan/skills` do not carry it. The string was evidently stamped by the same batch pipeline for both families (see #770 for where the caveman stamp is written).
2. **[LOW]** Because `translator` is what `tools/refresh-untranslated-stubs.mjs` reads to decide whether a mirror is a stub (#798), the mislabel is harmless to that tool — anything other than the stub literal is refused — but any future attribution query over the corpus (#769) would count 1,041 wenyan files as caveman work.

## Acceptance Criteria

- [ ] Decide the wenyan attribution string (the homage names the caveman spellbook's inspiration; whether wenyan shares it or carries its own is a content decision, not a tooling one) and record it in `i18n/README.md` beside the caveman one
- [ ] Rewrite `translator:` in the 1,041 wenyan-tier files to that value in one commit that touches nothing else, verified by the same grep reading 0 afterwards and `node tools/translator-stamp.mjs --verify` staying green
- [ ] The scaffold or compression step that writes the string (#770) writes the per-family value, so the next wenyan batch does not reintroduce the caveman string

## Context

Measured 2026-09-07 in the worktree at `0c8d5afbc` (PR #800, fact F14). Related: #770 (the homage stamp is written on byte copies), #769 (what `translator:` may say for an unreviewed machine translation).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #785, opened 2026-09-03
A7c scope: skill path references are unchecked in the ten i18n mirrors, inside code fences, and under every prefix but three


pjt222:
## Summary

Integrity check A7c (`scripts/check-skill-path-refs.js`, #773) resolves backticked repository paths a skill names, and its scope is deliberately narrow: **inline** references, in the **English** `skills/<id>/SKILL.md`, under **three** prefixes. That matched #773's acceptance criterion. The complement is not covered by anything, and this repository's standing rule is that an English-side edit multiplies by ten, so the i18n half is the one that grows silently.

## Findings

1. **[MED]** The ten `i18n/<locale>/skills/<id>/SKILL.md` mirrors carry the same references and are scanned by nothing. A path renamed in English and propagated to the mirrors is fine; a path renamed and *not* propagated leaves ten dangling references that no gate sees. Note the interaction with the frozen-fence rule: a reference inside a code fence is frozen and must match English byte for byte, so the mirrors' fenced references are covered transitively — the inline ones are not.
2. **[MED]** Paths inside fenced code blocks are outside the predicate in English too. `extractRefs` matches inline backticks only, so a `bash` fence naming `scripts/gone.js` resolves to nothing and is invisible.
3. **[LOW]** Only `workflows/`, `tools/` and `scripts/` are checked. A skill naming `guides/x.md`, `agents/y.md`, `teams/z.md`, `skills/<id>/SKILL.md` or `.github/workflows/w.yml` is unchecked, and the first four are the cross-reference paths the repository's own conventions ask skills to carry.
4. **[LOW]** The dot-in-last-segment rule excludes every extensionless file, so `tools/Makefile` or `scripts/LICENSE` would be silently skipped rather than resolved. Documented in the module header as the price of the rule; a `git ls-files` membership test would remove the compromise.
5. **[NOTE]** Agents, teams and guides carry repository paths too and are scanned by nothing at all.

## Acceptance Criteria

- [ ] Decide and record which of (1)–(5) A7c should cover; each widening is a separate, measured step with its own corpus run, not one sweep
- [ ] For the i18n half, the check runs per locale and reports the locale in the finding, so a mirror-only miss is distinguishable from an English one
- [ ] Any widening keeps the exact-set allowlist property in both directions (a waived path that starts existing, and an entry no file carries, are both failures) and keeps the vacuous-scan refusal (zero references scanned is exit 2)
[...]

PR #854, opened 2026-09-16
fix(skills): the four redaction skills name tooling that does not ship (#751)

pjt222:
## Summary

Part of W3-P3 (#751). Three of the four redaction skills — `enforce-redaction-gate`, `redact-wire-capture`, `redact-visualization-for-disclosure` — present invocations of `tools/enforce-redaction-gate.sh` and `tools/redact-visualization.py` as runnable commands, though neither script has ever existed in this repository. `redact-for-public-disclosure` is the one case where the named script (`tools/check-redaction.sh`) genuinely ships.

**Round 1 review found the original ruling rested on a false premise** (mine, not the reviewer's execution — see round-1 notes below). Round 2 is now in progress.

## What changed

**Fence bodies are corrected directly**, in the three skills that name the two absent scripts. Verified first: only `redact-for-public-disclosure` has i18n mirrors (10/10 locales); the other three have zero (checked across all 10 `i18n/<locale>/` directories). With no mirrors to orphan, the corrections live as a comment *inside* each fence — reaching a reader who searches or greps the invocation, not only one reading top to bottom — rather than as prose above it. `redact-for-public-disclosure`'s fences are untouched: its script genuinely ships and it genuinely has mirrors, so editing inside its fences would require propagating to all 10 — out of scope (#631, #852).

**`enforce-redaction-gate` Step 2's own scanner sketch failed open.** A nonexistent or unreadable `TARGET` produced output byte-identical to a clean tree — no output, exit 0 — because `2>/dev/null` swallowed the tool error and `set -uo pipefail` without `set -e` let `LEAKS` stay 0. Verified by extracting the sketch's actual bytes from the file and running them against a missing path. Fixed with an explicit fail-closed check plus the real 0/1/2 contract (verified against the file's own bytes: clean=0, missing/unreadable=2, leaky=1 regardless of count), replacing the "exit code = leak count" design the skill condemns everywhere else. `tools/check-redaction.sh` is now cross-referenced from all four skills as the shipped reference implementation — via the in-fence corrections themselves.

**`redact-for-public-disclosure`**: the real, shipped exit contract of `tools/check-redaction.sh` stated as prose below Step 7's fence, read from the script itself — **0 clean / 1 any findings (count in the printed output, never the exit status) / 2 could not run**, "2 must never be read as a pass" stated explicitly. The reasoning cites the two failure modes that are actually reachable in a 3-pattern sketch — the exit-2 collision (measured: three patterns, two seeded leaks, `exit=2`) and the missing-reserved-state problem that lets a `scanner && ok || echo CLEAN` wrapper read a crash as clean — not the 256-finding wrap the sketch's own pattern count can never reach. Step 7's "design choices" bullet reverted to "non-zero blocks the workflow" (not "1 or 2") — the CI step in that skill's own narrative runs the *reader's* Step-3-built scanner, a raw count 0-3 by the sketch's own design, not this repository's real script.
- `REDACTION_POLICY.md`, which this repository also does not carry, softened to "each consumer supplies" at the two places the skill's Validation/Expected blocks assumed it existed.
- Four version bumps (1.1→1.2 / 1.0→1.1) and no `skills/_registry.yml` changes — the four descriptions are reverted to their original text. A registry description is catalog/viz text with no referent for "this repository ships no X"; that claim belongs in the SKILL.md body only, where it already lives via the fence corrections. Confirmed byte-identical to `origin/main`, and confirmed `viz/public/data/skills.json` needs no regeneration from this PR — it's already stale for an unrelated, pre-existing reason (agent count 75→76), out of scope here.
- No `scripts/check-skill-path-refs.js` changes either. Round 1 added four `ALLOWLIST` entries to waive backtick-quoted mentions of the absent scripts that briefly existed in prose above the fences; that prose is gone now (moved inside the fences as plain comments, which the gate has never scanned), so the entries went stale and are reverted along with the file.

## Weakest claim

Three skills still functionally depend on a script (`tools/enforce-redaction-gate.sh`) nobody has decided to write or formally drop — this PR narrates that gap accurately now, in a place readers actually land on, but doesn't close it (#853). And `check-skill-path-refs.js`'s A7c gate (#773) still cannot see any of this: it only matches backtick-quoted paths, so a bare invocation inside a fence — exactly what all five corrected call sites are — stays invisible to it. That's the detector-blind-class finding, already covered by #785 (finding 2), not re-filed; measured evidence added as a comment there.

**#751 is not closed.** It has six acceptance criteria; this PR now satisfies two in full and touches a third partially:
- **Satisfied**: "Each skill states which of its named scripts ship... so 'the gate' is never read as something already running" (the fence corrections). "`tools/check-redaction.sh` cross-referenced from all four skills as the shipped implementation" — now true via the in-fence corrections (all three skills that named the absent gate now also name `check-redaction.sh` as the shape-tier half that does ship).
- **Touched, not fully closed**: "`redact-for-public-disclosure` Step 2 states a default for the no-policy-adopted case, and addresses material already published by a third party" — the REDACTION_POLICY.md softening states it's per-consumer, but doesn't add a stated default or address third-party-published material.
- **Untouched**: "`enforce-redaction-gate` Step 3 gains a tier/caveat for an internal identifier in prose/comment position" and "the Validation checklist gains a canary check for that gap" — genuinely separate defect (the structure-aware tier's blind spot, not the tooling-existence problem this PR addresses), out of scope here. "Consider whether `REDACTION_POLICY.md` should exist for this repo" — touched by the softening but no explicit decision recorded.
- #751's own finding 2 names six absent artifacts (`check-redaction.sh`, `enforce-redaction-gate.sh`, `sync-to-public.sh`, `redact-visualization.py`, `public-allowlist.txt`, `REDACTION_POLICY.md`). This PR states four of the six (the two it doesn't touch, `sync-to-public.sh` and `public-allowlist.txt`, are pre-existing `ALLOWLIST` entries in `check-skill-path-refs.js` predating this PR, and `public-allowlist.txt` is asserted as existing in that file's own Validation checklist) — a scoping call, not silent: those two are `redact-for-public-disclosure`'s own file-sync tooling, a different mechanism from the two-tier gate this PR is about, and left for a separate pass.

Filed as #852 (fence residual, blocked on #631) and #853 (operator decision). Neither closed by this PR, and neither is #751.

## Verification

```
$ npm run check-readmes
...
All files are up to date.

$ node scripts/check-i18n-fence-parity.js --locale de --id redact-for-public-disclosure
i18n fence parity: 7 fences in 1 translated skills
compared against every historical revision of their English source.
OK: every gated code fence matches an English source revision.
(unaffected — this skill's fences were never touched)

$ npm run test:scripts
...
ℹ tests 862
ℹ pass 862
ℹ fail 0

$ npm run validate:integrity
...
=== Summary ===
PASSED: All integrity checks passed
WARNINGS: 7 warning(s) detected
(same 7 pre-existing, unrelated glyph/palette warnings as round 1)

$ git diff $(git merge-base HEAD origin/main) --stat -- skills/ scripts/
 skills/enforce-redaction-gate/SKILL.md              | 18 ++++++++++++------
 skills/redact-for-public-disclosure/SKILL.md        | 10 ++++++----
 skills/redact-visualization-for-disclosure/SKILL.md |  8 +++++++-
 skills/redact-wire-capture/SKILL.md                 |  5 ++++-
 4 files changed, 29 insertions(+), 12 deletions(-)
(the true PR diff against the actual merge-base — origin/main has since advanced with unrelated merges)
```

All four `SKILL.md` files stay well under the 500-line limit, and under the tighter 494 ceiling that applies once a mirror exists (#855) — though only `redact-for-public-disclosure` has mirrors: 316 / 217 / 137 / 161 lines.

## Round 2

Head now `e0dfe1e68870a488b80723c2ac7b9c33096b2b22`. Round 1's fix closed only the **target-side** fail-open hole (`[ -r "$TARGET" ]`). The **scanner-side** hole was untouched: `rg`/`grep` erroring mid-scan — on a leaky tree or a clean one — produced byte-identical output (nothing) because `2>/dev/null` swallowed the error and `set -uo pipefail` without `set -e` left `LEAKS` at 0 either way. Fixed: the search tool's real exit status now survives `| head -1` under the sketch's existing `pipefail`, captured as `rc=$?` immediately and checked against a `<= 1` threshold (0 = match, 1 = rg/grep's own no-match convention, anything higher is a genuine tool error) — fail closed, exit 2, never a false clean. Added a `command -v "$SEARCH"` check too, since the grep fallback was never itself confirmed present.

All five arms — clean, missing target, leaky, scanner-error-on-clean, scanner-error-on-leaky — verified against the file's own extracted fence bytes (not a hand-typed copy), on both the `rg` and `grep` code paths, and `bash -n`/`zsh -n` clean, before committing.

Four prose fixes from the same round: the Validation checklist now says "1 specifically" (exit 2 previously also satisfied "non-zero"); Step 7's opener in `redact-for-public-disclosure` now says "This repository's `tools/check-redaction.sh`"; the five in-fence comments no longer claim the absent scripts don't exist "in this repository or any other" (unmeasurable, and false the moment someone builds one); `redact-visualization.py`'s comment now cites #852, not #853 (whose ACs only cover `enforce-redaction-gate.sh`).

Round 1's `ALLOWLIST` revert was independently verified correct, but traded away the one mechanism that would fire the day #853 resolves toward writing the file (`present && waived`). Restored via the cheaper option: one backtick-quoted path per skill in that fence's own Expected block (a forward-looking "not yet checkable" clause, not a restatement), with four `ALLOWLIST` entries whose reasons no longer claim existence anywhere. `node --test scripts/test/skill-path-refs.test.js`: 10/10, including the exact-set and reason-shape assertions.

Verified again: `npm run test:scripts` (862/862), `npm run validate:integrity` (PASSED, same 7 pre-existing warnings), `npm run check-readmes` (clean), fence parity for `redact-for-public-disclosure` (clean, unaffected), line counts 316/219/137/161 (all under 494), true PR diff against the merge-base scoped to the four `SKILL.md` files plus `scripts/check-skill-path-refs.js` (50 insertions / 19 deletions across round 2).

## Related

Part of #751 (not closed, see above). Blocked follow-up filed as #852 (references #631). Operator decision filed as #853. Comment with measured evidence added to #785. Also found and reported separately (not part of this PR): a translator-field mislabel on this skill's wenyan mirrors, already tracked as #801 (commented with a narrower, additional finding — only the `description:` field lags English, the body is genuinely translated).


pjt222:
## Adversarial round (Opus): one blocking finding, and it is mine not this PR's

Round run against the bundle cut at `44ab6e8a7`, `BUNDLE_STATUS` clean. I asked the reviewer to attack my own ruling first, and it did.

### BLOCKING — three of these four skills have zero i18n mirrors

I ruled that fence bodies must stay byte-identical because editing them "multiplies by ten mirrors with no tool to propagate into translated ones". That shaped this entire PR. It is false for every file it shaped. Re-derived by me at `333438467` after the reviewer flagged it:

| skill | mirrors | script it names in a fence |
|---|---|---|
| `redact-for-public-disclosure` | **10** | `tools/check-redaction.sh` — **ships** |
| `enforce-redaction-gate` | **0** | `tools/enforce-redaction-gate.sh` — absent |
| `redact-wire-capture` | **0** | `tools/enforce-redaction-gate.sh` — absent |
| `redact-visualization-for-disclosure` | **0** | `redact-visualization.py`, `enforce-redaction-gate.sh` — absent |

Exactly inverted from what I assumed. The only skill with mirrors is the only one whose named script ships — the file #751 never asked anyone to change. Every fence carrying a never-existent script sits in a file with no mirror to orphan. Of all skill directories, four lack a `de` mirror and three are these.

**The tell was in the plan row I quoted, verbatim, in the instruction:** `[DEBT] rfpd's 10 translated mirrors` — *rfpd only*, one skill of four. I read it, passed it on, and generalised "ten mirrors" to all four without listing a single `i18n/` directory.

There is a measurement that would have caught it and this PR did not run. `node scripts/check-i18n-fence-parity.js --locale de --id enforce-redaction-gate` exits **2**:

```
A typo and a real but untranslated id land here alike; both mean this run examines nothing.
```

The PR's cited parity check ran against `redact-for-public-disclosure` — the one skill where it had something to compare, and the one skill this PR is not really about.

**Corrected ruling: edit the five fence bodies directly in the three mirror-free skills.** Prose above a fence reaches a top-to-bottom reader and misses the one who lands on the fence by search, and the agent grepping for the invocation. With mirrors at zero, a correction *inside* the fence reaches both, for free. `redact-for-public-disclosure` stays fence-untouched — it genuinely has ten mirrors and its script genuinely ships.

**#852 has been corrected in place.** It repeated the false premise in its summary and carried an AC that cannot be satisfied — "propagate to all translated mirrors … verified by `check:fence-propagation --id <skill-id>` for each of the three" — against ids with no mirrors. The #631 blocker is removed; #631 is in any case a triage pass over the *detector*, not a repair tool, which was a second thing I had wrong and is corrected on #845.

### Substantive findings to fix in this PR

**2 (HIGH) — `enforce-redaction-gate` Step 2 still prescribes the design this PR condemns elsewhere, and its sketch fails open.** The heading `(exit code = leak count)`, the prose "The exit code equals the number of leaking shapes… what makes it composable", and its Expected block are untouched. `tools/check-redaction.sh`'s own header says that spec is unsound and names **both** skills. And measured on the sketch verbatim: a leaky tree with the scanner erroring and a clean tree are **byte-identical** — no output, exit 0 — because `2>/dev/null` swallows the error and `set -uo pipefail` without `set -e` leaves `LEAKS` at 0. That is the trap the third exit state exists for, reproduced in the code the skill tells the reader to write. All outside fences; and with no mirrors, even `exit "$LEAKS"` is free to fix.

**3a (MED-HIGH) — one sentence went from correct to incorrect.** "The job exits with the scanner's exit code; non-zero blocks the workflow" became "…any non-zero (**1 or 2**) blocks the workflow". The scanner that CI step runs is the Step 3 sketch, which exits with a count in `0..3`. `1 or 2` describes agent-almanac's script, not the one the step runs.

**3b (MED) — the registry description picks the wrong referent.** It now says `check-redaction.sh` "ships here" while the skill it indexes says the script lives in the reader's private repo. Separately, a `description` is user-facing catalog text rendered into the viz graph; "this repository ships no `tools/enforce-redaction-gate.sh`" is repository errata, and a graph reader has no referent for "here".

**4 (MED) — the imported rationale is the one that cannot occur.** `check-redaction.sh`'s header records two measured failure modes of count-as-exit: 256 wrapping to 0, and 2 colliding with the reserved could-not-run code. This PR quotes the first. But both sketches increment **once per pattern**, so `LEAKS` is bounded by the array — 3 and 4. 256 is structurally unreachable in the code the paragraph annotates; the collision at 2 fires on the second leak (measured: three patterns, two seeded leaks, `exit=2`). A reader checking the reason against the sketch concludes "256 is impossible, `exit $LEAKS` is fine" and keeps the design.

**5 (MED) — the "still open" list under-claims.** #751 has six ACs; two more are untouched and unnamed. **AC2**: a Validation check that fails when a real internal name in a comment is not caught — no Validation checklist in any changed file. **AC5**: `check-redaction.sh` cross-referenced from **all four** skills — currently 1 of 4, and adding it to `enforce-redaction-gate` Step 2 also discharges finding 2. Also, #751 finding 2 names six absent artifacts and this PR states three; `tools/sync-to-public.sh` and `tools/public-allowlist.txt` get no statement, and the latter is asserted as existing in that file's own Validation checklist.

**6 (LOW-MED) — `viz/public/data/skills.json` not regenerated** in the same commit as the registry change, per CLAUDE.md § Viz Deploy Model. Bounded: `deploy-pages.yml` regenerates on push, so the drift is in-tree only.

**7 (LOW) — two allowlist reason strings overstate.** The mechanism itself graded **correct**: keyed `<skill-id>:<path>` so it is not over-broad, exact in both directions so a resolved #853 forces a same-commit edit, and dead-entry detection keyed off the backtick the new prose introduces — so the pair is forced both ways. But two reasons are byte-identical and read as an existence claim ("the target project's gate script this skill verifies through"), which is the euphemism that file's own comment warns against.

### What passed, verified independently

The reviewer reverse-applied the diff and compared fences pre/post with its own extractor: **count, info-string tags, closed-ness and body SHA-256 identical across all four skills**. The PR's central mechanical claim holds. Line counts 215/316/159/136, the largest 178 below the **494** ceiling (#855). Version bumps right — no step removed. The exit contract as stated is faithful to the shipped script, verified from source.

### The framing question, answered empirically

I asked whether reframing a fence in prose above it is sufficient or a documentation fix to what a reader experiences as a broken command. The reviewer's answer: the trade-off never existed for these three files, so it resolves by measurement rather than argument. That is the better answer, and it only became available because someone checked the premise instead of debating it.

pjt222:
## Round 2 (Opus) at `33d64b944`: round 1's blocker is fully resolved; one HIGH remains

Same reviewer, bundle cut at `33d64b944` with `diff-since.md` from the head it graded in round 1. Its disposition: **nothing blocks in the round-1 sense** — every checkable claim in the PR body is true, and the change is better than both `main` and the previous head. One finding should still land before merge, because newly written prose asserts a property the code below it does not have.

### The one to fix: the fail-closed repair closed the wrong hole

Step 2 gained `[ -r "$TARGET" ] || { …; exit 2; }`. That closes the **target-side** failure. Round 1's finding was the **scanner-side** one, and it is untouched. Measured on the fence bytes extracted from the committed file:

```
clean tree, rg present               exit=0   (no output)
missing target                       exit=2   cannot read: ./does-not-exist
leaky tree (3 shapes)                exit=1   LEAK: … | … | …
rg+grep erroring, LEAKY tree         exit=0   (no output)
rg+grep erroring, CLEAN tree         exit=0   (no output)
```

The last two are **byte-identical** on a tree that leaks three shapes. `2>/dev/null` still swallows the search tool's error, `hit` stays empty, `LEAKS` never increments — and the new explicit `exit 0` now certifies the leaking tree as clean.

What makes this a defect in *this* PR rather than a leftover is that the new prose claims the opposite: *"a genuinely crashed scanner [has] nowhere to signal from but 'zero'"* is offered as the problem being solved, and under the new design a crashed scanner still has nowhere to signal from but zero. Adding a `2` state to a mapping the crash never reaches changes nothing. The new Expected block is correspondingly narrower than the prose — it covers the arm that works and is silent on the arm that does not, so a reader who verifies it concludes the gate is fail-closed.

Not theoretical: the skill's own On-failure contemplates `rg` being unavailable, and `SEARCH=$(command -v rg >/dev/null && echo rg || echo grep)` selects `grep` without checking that `grep` exists either.

**The repair needs no new machinery** — the sketch already sets `pipefail`, so the search tool's status survives the pipe. `; rc=$?` after each `hit=` line, `[ "$rc" -le 1 ] || { …; exit 2; }` after each, and one `command -v "$SEARCH"` check. The reviewer measured the patched copy parsing under `bash -n` and `zsh -n` and giving all five arms, including `scanner erroring, LEAKY tree → exit=2`. These fences have no mirrors, so it costs nothing.

### Also worth landing, all cheap prose

- **The Validation checklist under-specifies the contract the skill just adopted.** It still says the gate exits *"non-zero on a tree seeded with a deliberate test token"* — under 0/1/2, **exit 2 "could not run" satisfies that checkbox**, so ticking it confirms nothing about the distinction the whole change exists to make. Mirror the Expected block. (The other two skills' Validation lines were checked and remain correct.)
- **The referent collision survives in the one place it was sharpest.** Step 3 was fixed to *"this repository's actual `tools/check-redaction.sh`"*; Step 7's opener did not get the same three words, and it sits directly below the fence that fetches the **reader's** copy from their private repo. Apply Step 3's own repair.
- **Five in-fence comments claim the script "does not exist in this repository or any other yet."** #751's evidence was scoped to this repository and #853's summary says *"in this repository"*. "Or any other" is unmeasurable, and it is false the moment the reader builds the file the surrounding sentence tells them to build.
- **`tools/redact-visualization.py`'s comment cites #853**, whose scope is `enforce-redaction-gate.sh` only — every AC there names that file. **#852** is the issue that covers the `.py`; its finding 4 is exactly that call site.
- **Three PR-body imprecisions**: AC3 is labelled flat "Satisfied" while the body itself discloses the `sync-to-public.sh` / `public-allowlist.txt` pair has no statement, and the AC says *each*; AC6 is under-claimed (the softening to "or equivalent" does satisfy its second branch); and AC5 is satisfied but not by the stated route — for `enforce-redaction-gate` it is Step 2's **prose** that names `check-redaction.sh`, not either fence comment.

### The trade nobody named: reverting the gate file gave up the only tripwire

The mechanical claim **checks out** — the reviewer imported the reverted `check-skill-path-refs.js` at the bundle sha and ran its real `extractRefs`/`checkSkill`/`ALLOWLIST` over the four post-change files: no findings, no stale entries, no dead entries. The revert is correct.

But the same probe measures what it cost:

```
mentions of the two absent scripts across the three skills:  8 lines
of which backticked (i.e. visible to the checker):           0
```

Round 1's four allowlist entries had a `present && waived` rule that was **the only thing in this repository that would fire the day `tools/enforce-redaction-gate.sh` gets written** — which is precisely what #853 exists to decide. Five newly written factual assertions now have no owner that fails, in a repository whose standing rule is that a prose claim needs a generator or a pinned assertion.

Relative to `main` this is not a regression; the paths were unbackticked inside fences there too. Relative to round 1 it is a trade, and the PR body describes it passively as though nothing was given up. Two cheap ways to keep both: **one** backticked mention per skill in prose (an Expected block is the natural home) so the entries stay live as tripwires while the in-fence comments do the reader-facing work, or a named member list in `debt-ratchet.yml` with #853 as the exit.

### Verified independently — what the rework got right

- **Fence scope is exactly right.** Every fence in all four files compared against `origin/main` by tag, body SHA-256, count and closed-ness. `redact-for-public-disclosure`: **all 7 fences byte-identical to `origin/main`** — its ten mirrors are untouched, as claimed. The three mirror-free skills changed exactly the five intended fences; every untouched fence in those files is byte-identical.
- **All 13 `bash` fences parse under `bash -n` AND `zsh -n`.**
- Round-1 findings 3a, 4, 5, 6 and 7 are resolved or moot, each checked rather than accepted. Finding 1 fully resolved.
- The `skills.json` and registry-revert claims hold.

Handing back to the author for finding 1 and the prose items. **Not merging** until those land.

pjt222:
## Round 2's HIGH re-derived independently and confirmed fixed — merging

Not spending a third round: round 2's verdict was already *"nothing blocks in the round-1 sense"*, the one HIGH was specified precisely by the reviewer, and I re-derived the repair myself against the bytes extracted from the committed file at `e0dfe1e68` rather than taking the report's word for it.

All eight arms, run against the extracted Step 2 fence:

```
clean tree                              exit=0
leaky tree, 1 match                     exit=1   LEAK: vendor flag namespace
leaky tree, 400 matches                 exit=1   LEAK: vendor flag namespace
missing target                          exit=2   cannot read: …/nope
unreadable subdir, LEAKY tree           exit=2   cannot scan: …/denied (tool error)
scanner erroring, CLEAN tree            exit=2   cannot scan: …/clean (tool error)
scanner erroring, LEAKY tree            exit=2   cannot scan: …/leaky1 (tool error)
no scanner on PATH at all               exit=2   cannot find grep
```

The two that matter are the last rows of each pair. `scanner erroring, LEAKY tree` was **byte-identical to a clean run at exit 0** before this fix — a gate certifying a tree with three leaks. It now exits 2, matching the reviewer's measurement exactly. And `unreadable subdir, LEAKY` is the realistic form of the same failure: a permission-denied directory inside the scanned tree, which is how this actually bites.

**Two arms I added that were not in the spec**, both clean:

- **400 matching files.** `| head -1` closes the pipe early, so I expected `$SEARCH` to take a SIGPIPE and exit 141 — which `[ "$rc" -le 1 ]` would read as a scan failure and fail closed on a merely-very-leaky tree. It does not happen: exit 1 with the leak reported. Worth having measured rather than assumed, because `pipefail` plus an early-closing consumer is exactly where that goes wrong.
- **No scanner on PATH at all.** My first attempt at this set `PATH` to one empty directory and read `exit=127` — but that also removed `head`, so the gate died inside the loop and the reading said nothing about the `command -v "$SEARCH"` check. That was my harness, not the gate. Re-run with a PATH carrying coreutils and omitting only the search tools: `exit=2, cannot find grep`, on both clean and leaky trees.

`bash -n` and `zsh -n` both clean on the published fence.

**On item 7**, the option taken is the better of the two I offered: one backtick-quoted path per skill in that fence's own Expected block, phrased as a forward-looking "not yet checkable" clause rather than a restatement of the in-fence comment. That keeps the four allowlist entries live as tripwires for the day #853 resolves as "write it" — which was the only thing in the repository that would fire then — while the in-fence comments do the reader-facing work, and it adds information instead of only re-triggering the gate.

15/15 contexts green at `e0dfe1e68`. Two adversarial rounds on this PR, both acted on in full. Merging.

For the record, since it shaped this PR from the start: round 1's blocking finding was **mine** — I ruled that fence bodies could not be edited "because ten mirrors", when three of these four skills have zero. The author executed a wrong instruction correctly, then reworked the whole approach when it was reversed, and pushed back correctly three times on things I had wrong (the true merge-base, the unrelated `skills.json` drift, and #801 already covering the wenyan mirror).

pjt222:
## Correction: my verification of the SIGPIPE arm covered one of its two branches. Filed as #858.

In the merge comment above I wrote:

> `leaky tree, 400 matches   exit=1` … "`| head -1` closes the pipe early, so I expected `$SEARCH` to take a SIGPIPE and exit 141 — which `[ "$rc" -le 1 ]` would read as a scan failure and fail closed on a merely-very-leaky tree. **It does not happen**"

It does happen. My probe ran with `rg` on `PATH`, so it exercised only the `rg` branch. Forced onto the `grep` fallback, the same fence on `origin/main`:

```
WITH rg present:                       FORCED onto grep:
  clean tree            exit=0           clean tree            exit=0
  leaky, 1 match        exit=1           leaky, 1 match        exit=1
  leaky, 500 matches    exit=1           leaky, 500 matches    exit=2   <-- WRONG
```

GNU grep 3.11 takes SIGPIPE from `head -1` and returns **141**, three runs of three, but only once there are enough matches to fill the pipe buffer — one match returns 0. `rg` does not take it, which is why the branch I tested was clean and why the prediction read as refuted.

**Severity is low and the direction is safe:** this is fail-CLOSED. The gate refuses rather than certifying a leaking tree, so it is not a security hole — it is the gate becoming unusable on a machine without `rg` exactly when a tree has enough leaks to matter. Strictly better than the fail-OPEN this PR removed, and not a reason to revert anything here.

The fail-closed guard itself (`rc=$?` plus `[ "$rc" -le 1 ]`) is correct and stays. `| head -1` is what makes it misfire, and it buys nothing — the sketch only asks *whether* a pattern matched, never which file. One deletion per branch; measured, `rc=0` on 500 matches without it.

Tracked as **#858** rather than reopening, since this merged and the fix is a separate one-line change.

Found by a design agent building `tools/enforce-redaction-gate.sh` for #853, which measured the same SIGPIPE at 3/3 and removed `| head` from its own design before anyone asked. Re-derived against the merged fence before filing.

**The transferable part**, because it is the fourth instrument failure I have logged today and the second to reach something published: a probe that exercises one branch of a two-branch fallback has verified one branch. I predicted the exact failure, wrote a test for it, and reported it absent — the prediction was right and the measurement was aimed at the wrong half.

Issue #855, opened 2026-09-16
[MED] The 494-line English ceiling is a prose constant with no owner, and the local scan misses new skills


pjt222:
## Summary

`CLAUDE.md` § Skill Validation now says: *"For a skill with scaffolded mirrors the English ceiling is **494**."* The arithmetic is correct today and was checked against real files — English 491, mirrors 497, delta exactly 6 — but **nothing computes or asserts 494**. It is a number in a paragraph, which is the defect class this repository already has a name for: a prose count needs an owner that fails.

Found by the adversarial round on #843 (finding F4), which is the PR that introduced the number. Filing rather than bolting a fix onto that PR, because the durable fix is a check and not a better sentence.

## Findings

1. **[MED]** 494 has no owner. The ceiling is `500 − (provenance field count)`. A seventh provenance field silently makes it 493 and the sentence becomes wrong with nothing going red. The six fields are `locale`, `source_locale`, `source_commit`, `fence_basis_commit`, `translator`, `translation_date`.

2. **[MED]** The overhead is **not** a constant across the corpus, so a single number is wrong for part of it already. `CLAUDE.md` records 229 mirrors without `fence_basis_commit`, and `tools/refresh-untranslated-stubs.mjs` *drops* that field from any stub it rewrites until `--stamp` runs — so during that window the overhead is 5 and the ceiling is 495, not 494. A file at 495 would be legal by the true rule and illegal by the published one.

3. **[MED]** The local pre-check added in #843 does not protect the case that caused #843. It scans `skills` and `i18n` together, so it catches an over-length mirror **once the mirror carries the new body**. For an *existing* stub refreshed in the same branch — what happened in #843 — that works. For a **new** skill it does not: `CLAUDE.md` § Adding a New Skill says to scaffold "after the English is committed and final… on an external PR that means after merge". At the moment the local scan runs, the mirrors do not exist or carry an older body, the scan is green, and the mirrors land at English + 6 on `main`. For that path 494 is the only protection and it is prose.

4. **[LOW]** The same paragraph is the only statement of the rule. A contributor who reads `SKILL.md files must stay under 500 lines` — which appears earlier in the same bullet and in `skills/create-skill/SKILL.md` — and stops there gets the wrong ceiling with no signal.

## Acceptance Criteria

- [ ] The English ceiling is **derived**, not written down: a check computes `500 − (number of provenance fields a scaffolded mirror of THIS skill would carry)` per English file and fails when the file exceeds it
[...]

Issue #660, opened 2026-08-18
skills: a scope declaration stored in CONTINUE_HERE.md is destroyed by the first reader


pjt222:
## Summary

`coordinate-peer-sessions` offers `CONTINUE_HERE.md` as one of three places to record a path-and-branch
scope declaration. `read-continue-here` deletes that file after one consumption, by design. So the
declaration survives exactly one reader — and nothing in either skill carves out an exception for
content that is not a handoff.

Found by adversarial review of PR #659 as a secondary observation, not a defect in that diff.

## Findings

1. **[MEDIUM]** The two skills disagree about the file's lifetime.

   `skills/coordinate-peer-sessions/SKILL.md:105-106` — Step 3, where the declaration is written:

   > Record it where the other session can read it — a message to the human running both, a line
   > in `CONTINUE_HERE.md`, or a comment on the issue.

   `skills/read-continue-here/SKILL.md:116` — Step 5, "Clean Up":

   ```bash
   rm CONTINUE_HERE.md
   ```

   with the rationale "Stale continuation files cause confusion in future sessions." That rationale is
   correct *for a handoff* and wrong for a standing agreement: a scope declaration is not stale when it
   is read, it is in force. Whichever session reads first deletes the other's declaration, and the
   deleting session has no way to tell the two kinds of content apart.

2. **[MEDIUM]** The optional SessionStart hook makes this the default path rather than an edge case.
   `read-continue-here` Step 6 configures a hook that injects the file into **any** session starting in
   the project. With it enabled, a peer session consumes the declaration on startup — and then the
   deletion in Step 5 is the documented follow-up.

3. **[LOW]** The file does not travel, which the declaration text does not say. It is gitignored here:

   ```console
   $ git check-ignore -v CONTINUE_HERE.md
   .gitignore:22:CONTINUE_HERE.md	CONTINUE_HERE.md
   ```

   So it is a same-checkout channel only — fine for a peer sharing the worktree, which is the case
   `coordinate-peer-sessions` addresses, but it means the third option ("a comment on the issue") is the
   only one of the three that survives a clone.

## Acceptance criteria

- [ ] One of: (a) `coordinate-peer-sessions` Step 3 stops offering `CONTINUE_HERE.md` and names a
      durable channel instead, or (b) `read-continue-here` Step 5 is scoped so it deletes the handoff
      without deleting a standing declaration, or (c) a distinct file is named for declarations and both
      skills reference it
[...]

Issue #628, opened 2026-08-17
i18n: fence-delimiter corruption hides inside the unjudged 'unalignable' pool — 7 write-vignette mirrors carry a 4-backtick count English never had


pjt222:
## Summary

Repairing #598 finding 1 — three `escalate-issues` mirrors whose 4-backtick ` ````markdown `
opener had been degraded to three — raised the obvious question: was that three files, or a class?

It is a class, and the gate is structurally blind to it. **7 `write-vignette` mirrors carry a
4-backtick delimiter count that appears in NO English revision**, and every one of them is sitting
in the 60-file `unalignable` pool, which the gate deliberately does not judge.

## Why the gate cannot see it

A degraded opener usually changes the fence COUNT. `compareTagSequence` returns
`{ unalignable: true }` when no English revision has that count, and `unalignable` is reported as
unjudged rather than as a violation — correctly, because a translation predating a fence English
later gained lands there too, and calling that a violation reintroduces the staleness confound the
gate exists to avoid.

So the three that WERE caught were caught by luck: their degradation happened to preserve the fence
count, which is what let the tag-sequence check make a positional claim at all.

Nothing measures what is inside the `unalignable` pool. Its size is printed; its contents are not.

## The measurement

English 4-backtick delimiter counts across every revision, versus what each mirror carries:

| skill | English counts, ever | mirrors at a value English never had |
|---|---|---|
| `plan-release-cycle` | 0, 2 | 0 of 3 |
| `setup-putior-ci` | 0, 2 | 0 of 1 |
| `tidy-project-structure` | 0, 2 | 0 of 3 |
| **`write-vignette`** | **0, 6, 8** | **7 of 10** |

The seven: `caveman`, `caveman-lite`, `caveman-ultra`, `de`, `es`, `ja`, `wenyan-ultra` — all
carrying **2**, a count English went from 0 to 6 to 8 without ever passing through. The other three
(`wenyan`, `wenyan-lite`, `zh-CN`) carry 0 and are honestly stale.

Confirming one is unjudged rather than clean:

```bash
node scripts/check-i18n-fence-parity.js --locale de --id write-vignette --json
# tagSequenceUnalignable: 1, tagSequenceFindings: 0, tagSequenceDrift: 0
```

## A correction, because the first method was wrong

The first pass compared each mirror against English **at its own `source_commit`** and called any
difference a defect. That flagged 17 mirrors. It is the wrong question: `source_commit` is a claim,
[...]

Issue #468, opened 2026-07-30
skills: read/write-continue-here hardcode the project root, so a docs/CONTINUE_HERE.md layout silently no-ops


pjt222:
## Summary

`read-continue-here` and `write-continue-here` both hardcode the handoff file to the **project root**, with no way to configure a different location. A consuming project that keeps it elsewhere is silently unsupported: the SessionStart hook finds nothing and reports no error.

`skills/read-continue-here/SKILL.md`:

```
:28  - Starting a new session and CONTINUE_HERE.md exists in the project root
:43  Check for `CONTINUE_HERE.md` in the project root:
:46  ls -la CONTINUE_HERE.md 2>/dev/null
:113 After the handoff is consumed and work is underway, delete CONTINUE_HERE.md
```

and the SessionStart hook it installs:

```bash
CONTINUE_FILE="$PWD/CONTINUE_HERE.md"
if [ ! -f "$CONTINUE_FILE" ]; then
  exit 0
fi
```

`skills/write-continue-here/SKILL.md:58,94,96` mirrors it ("Write the file to the project root", "same directory as `.git/`").

## The concrete miss

`pjt222/bare-metal` keeps its handoff at **`docs/CONTINUE_HERE.md`** — it was moved out of the root during that repo's Tier-4 docs reorganisation (`CHANGELOG.md:157`), and its `CLAUDE.md` documents the `docs/` path as the convention. So:

- The installed hook's `exit 0` on missing file means **a wrong path is indistinguishable from "no handoff exists"**. There is no warning, no log line, nothing. Silent no-op is the worst possible failure for a session-bootstrap mechanism, because the symptom is "the agent didn't seem to know what I was working on" — which reads as a model problem, not a config problem.
- `read-continue-here` Step 5 runs `rm CONTINUE_HERE.md`. In a project using `docs/`, that either fails or, worse, deletes an unrelated root file if one exists.
- The agent works around it by hand every session (that is what happened in bare-metal on 2026-07-29: the skill was invoked, found nothing at the root, and the path had to be located manually).

## Second, smaller divergence

[...]

PR #856, opened 2026-09-16
fix(skills): state the handoff lifecycle instead of enforcing one, and stop routing peer scope through it (#775, #660)

### LINKED ISSUE #775: write-continue-here enforces gitignoring the handoff, which breaks the tracked-handoff lifecycle and its own delete-after-use rule

pjt222:
Closes #775 and #660. Completes #468's ACs 4 and 5 — #846 took ACs 1/2/3/6.

This is W3-P2 of the backlog plan, unblocked by #846's merge. Prose and one new detection fence; no tool, no gate.

## Premises

Each line names what this change depends on and where it was checked. Attack these first.

| This is correct only if… | Verified where | Result |
|---|---|---|
| `write-continue-here` really did *enforce* gitignoring, rather than merely suggest it | the two sites: Step 2's On-failure ("Verify `.gitignore` contains `CONTINUE_HERE*.md` … and if not, add it") and a Validation checkbox | Holds — an imperative and a box to tick |
| the two skills genuinely disagreed | `read-continue-here` Step 5 branches on `git ls-files --error-unmatch` and treats tracked and untracked as equally normal | Holds — the pair contradicted each other on whether a tracked handoff can exist |
| Step 5's citation pointed at nothing | read every occurrence of `track`, `ignore`, `gitignore` in `write-continue-here` before this change | Holds — "`write-continue-here` Step 2 states the trade-off" was true of no sentence in the file. Found by the #846 review, not by me |
| a scope declaration in a handoff is actually destroyed | `read-continue-here` Step 5 deletes the file it read; that is the skill's documented purpose | Holds — and it is the whole of #660 |
| the four `coordinate-peer-sessions` mirrors are stubs, so refreshing them is mechanical | `refresh:stubs --verify` before and after | Holds — 4 mirrors, 4 stubs, 4 diverged at line 17 (the version bump), 0 refused |
| adding a fence to `write-continue-here` does not move any gate | `check:fence-propagation --id write-continue-here` before and after | **It widens a known debt without changing a verdict** — see Known cost below. Stated rather than buried |

## What changed

**`write-continue-here` 1.2 → 1.3.** The two enforcement sites are gone. New **Step 4** states the trade-off and enforces nothing: what consuming costs under each lifecycle, what a stray `git add -A` does to each, and which kind of repository each suits. It *detects* the current state rather than assuming, and names a third the old text had no room for — **UNDECIDED**, untracked and not ignored, belonging to neither lifecycle and the state a `git add -A` can sweep up. Even there the instruction is to report it, not to edit anyone's `.gitignore`.

**`read-continue-here` 2.0 → 2.1.** One word: the Step 5 citation now points at Step 4, which exists and says what the sentence claims.

**`coordinate-peer-sessions` 1.0 → 1.1 and its guide.** Both offered `CONTINUE_HERE.md` as one of three channels for a path-scope declaration. It is the one that cannot work: a handoff is addressed to the *next* session and is consumed and deleted by the first reader, so the declaration is destroyed by the reader it was meant to bind while the peer that still needs it finds nothing. A declaration must outlive being read; a handoff is defined by not doing that. The Related Skills entry that advertised it as "one place a path-scope declaration can live" now says the opposite.

**This repository's `.gitignore`** keeps its rule and gains the reason — ignoring is *our* choice, with the cost stated (consuming is final, so an edition worth keeping goes to `misc/handoff-archive/<date>/` first), rather than something the skill demanded of us. That is the follow-up the plan's W3-P2 row names.

## Known cost, accepted deliberately

Step 4's detection block is a new frozen fence in a skill whose ten mirrors have been structurally misaligned since 1.1 (#845). English therefore goes to five fences against their two:

| at | frozen / total | mirrors alignable |
|---|---|---|
| `ee55ed509` | 2 of 3 | 0 of 10 |
| `84c331eec` (#846) | 3 of 4 | 0 of 10 |
| here | **4 of 5** | 0 of 10 |

No gate's verdict changes and no ratchet member moves, because the mirrors were already 0-of-10 alignable and `unalignable` is the unjudged pool (#628). The debt is now three fences deep rather than one, and that is the predicted price of sequencing #845 after this PR so the translation pass is paid once instead of twice. Recorded on #845.

Writing the detection as prose to dodge the fence gate was considered and rejected: a command the reader runs belongs in a fence, and changing punctuation to avoid a gate leaves the gate unable to see the class it exists for.

## Verification at `335330dd5`

```
npm run test:scripts                  875/875, 0 fail
npm run validate:integrity            PASSED (7 warnings, pre-existing glyph/palette)
npm run check-readmes                 All files are up to date.
check-i18n-fence-parity --locale de --id coordinate-peer-sessions
                                      OK: every gated code fence matches an English source revision
refresh:stubs --verify                4 mirror(s), 4 stub(s), 0 diverged, 0 without fence_basis_commit
```

Stubs refreshed, committed, then stamped at `d34f9ced6` in a second commit, per the standing two-step recipe. `translation_date` deliberately unchanged — for a stub it records the scaffold, not a translation event.

Line counts 217 / 307 / 317, all under the **494** ceiling that binds an English file with scaffolded mirrors (#855), not merely under 500.

## The weakest claim

**That UNDECIDED is worth a whole branch of the step.** I have measured it exists as a state; I have not measured how often it occurs in the wild, and the argument that it is "the one state worth acting on" is reasoning, not data. If it turns out to be vanishingly rare the step carries a branch nobody reaches.

Second weakest: #660's fix is prose in two files. Nothing *prevents* a session from writing its scope declaration into a handoff — the skill now advises against it, and advice is what the previous enforcement sites show is not the same as a mechanism. A gate here would need to understand what a scope declaration looks like, which is why option (a) was taken; if that is the wrong trade, this is the place to say so.


pjt222:
## Round 2 (Opus): two blocking findings, both fixed at `3122c1132`

Same reviewer as #846, which is why it went straight at the step rather than at the prose. Bundle cut at `335330dd5`, `BUNDLE_STATUS` clean. Both blockers re-derived independently before anything was changed.

### B1 — the step could not run in the case it was added for

Step 4's guard pointed at Step 2's resolver. That resolver assigns `CONTINUE_FILE` **only when a handoff already exists** — and Step 2's own third bullet is *"No handoff at all — this is the first one"*, where it is empty by design:

```
before Step 3 installs: CONTINUE_FILE=[]                  -> EMPTY, so the Step 4 guard aborts
after  Step 3 installs: CONTINUE_FILE=[CONTINUE_HERE.md]  -> set, Step 4 works
```

The remedy the guard offered was unsatisfiable — re-running the resolver returns empty again until Step 3's `mv -n`. And a project writing its **first** handoff is precisely the project that has not chosen a lifecycle, so the step aborted in its primary case. The sibling skill already carried the missing sentence (`read-continue-here` Step 2, *"shell state does not survive between steps"*); this one had no equivalent anywhere. Both fixed.

### B2 — `else` meant "everything I could not measure"

Seven cases against the extracted fence bytes, shipped against repaired:

| case | `ls-files` | `check-ignore` | shipped | repaired |
|---|---|---|---|---|
| tracked | 0 | 1 | TRACKED | TRACKED |
| ignored by the repo rule | 1 | 0 | IGNORED | IGNORED |
| untracked, not ignored | 1 | 1 | UNDECIDED | UNDECIDED |
| tracked **and** matching an ignore rule | 0 | 1 | TRACKED | TRACKED |
| **outside any repository** | **128** | **128** | **UNDECIDED** | NOT IN A REPOSITORY |
| **handoff in another repository** | **128** | **128** | **UNDECIDED** | IGNORED |
| **absent / unreadable `.git`** | **128** | **128** | **UNDECIDED** | NOT IN A REPOSITORY |

Exit 0 throughout, both stderr streams suppressed. The sharpest row is the first 128: **the step's own On-failure already knew the answer** — "outside a git repository the question does not arise" — while the fence printed "untracked and not ignored, the state most likely to be swept into an unrelated commit", where there is no repository and nothing that could sweep it.

The reviewer's framing of why this matters more here than in the sibling skill is worth keeping: `read-continue-here` Step 5 documents the same exit codes and deliberately does *not* separate them, and that is sound — its `else` performs one harmless action. Step 4's `else` **asserts a classification**. Conflating "not ignored" with "could not tell" is free when you are acting and wrong when you are reporting, and "it detects rather than assumes" was this step's whole claim.

Repaired by anchoring on the directory holding the handoff and probing for a work tree first.

### Also fixed, from the same round

- **`IGNORED` was not necessarily the *project's* choice.** A user-level `core.excludesFile` gives the identical verdict while a collaborator cloning the same repo gets UNDECIDED — one label, two facts, with "deletions are final" stated unconditionally. `check-ignore -v` now names the deciding file: `decided by: .gitignore:1:…` versus `decided by: /home/…/globalignore:1:…`.
- **A leading dash was parsed as options** — exit 129, `unknown switch 'C'`. Both commands now use `--`.
- **The `.gitignore` comment named `misc/handoff-archive/` repo-relative**, where it resolves to nothing; a session following it literally would create `agent-almanac/misc/`. It is one level above the repository and the comment now says so. My error, and I had used the correct parent-level path myself an hour earlier.
- **The glob rationale credited Step 2 with a rename that is Step 3's.**

### The apostrophe, which is the finding I'd keep if I could keep one

The first version of the repaired guard read `…or re-run Step 2's resolver…`. **An apostrophe inside `${VAR:?message}` is a bash syntax error and valid zsh:**

```
WITH an apostrophe:  bash exit=2  unexpected EOF while looking for matching `'
                     zsh  exit=1  X: re-run Step 2's resolver now that the file exists
WITHOUT it:          bash exit=1  X: re-run the Step 2 resolver now that the file exists
                     zsh  exit=1  X: re-run the Step 2 resolver now that the file exists
```

The `Bash` tool here runs zsh, so the zsh column is identical across both pairs and the broken form looked correct every way I could run it by hand. This is the **same class** the previous round found in `read-continue-here` — reintroduced the next day, by me, having written that finding up myself.

It was caught by `continue-here-blocks.test.js`, which runs the extracted fence under `bash -c`. That is the argument for the test rather than for reading more carefully: careful reading is what produced the bug.

### N4 — Step 4 now has an owner that fails

It was the only published block in this pair without one. New arm fixtures all four verdicts including both 128 cases. Three mutants, each **killed by exactly one test**:

| mutation | result |
|---|---|
| disable the work-tree probe (`if false;`) | KILLED by 1 |
| drop `-C "$HOME_DIR"` from the `check-ignore` arm | KILLED by 1 |
| delete the `decided by:` line | KILLED by 1 |

Two earlier attempts were **refused as no-ops** because the needles missed the file's exact bytes (`printf %s` against `printf '%s'`). That refusal is the tool working — a silently-unmatched mutation is what makes a negative test pass vacuously.

### On my stated weakest claim — the reviewer inverted it, correctly

I wrote that UNDECIDED might be "a branch nobody reaches". The reviewer's answer: it is not rare, it is **the default** — for a first handoff TRACKED is unreachable by construction, so a first handoff can only be IGNORED or UNDECIDED. The real problem was the opposite of the one I named: **over-capture**, the arm absorbing four states that were not undecided. Fix B2 and the branch becomes what I argued for; leave it and the branch's own text is the evidence against it. That is a better answer than either of mine.

### Verification at `3122c1132`

- `npm run test:scripts` — **876/876**, 0 fail
- `npm run validate:integrity` — **PASSED** (7 pre-existing glyph/palette warnings)
- `check:fence-propagation --id write-continue-here` — `4 frozen of 5 fence(s), 0 of 10 alignable`, unchanged in verdict, still #845
- Evidence: `tests/results/2026-09-16-lifecycle-step-classification/`

### Not taken, deliberately

The reviewer proposed a **cheap mechanism for #660** rather than advice alone: `coordinate-peer-sessions` emits a fixed `text` fence whose last line is `Nobody runs: …`, so a handoff containing that literal is carrying a scope declaration **by construction** — no understanding required — and `read-continue-here` Step 5 could warn before deleting. It pins a literal this repository controls rather than a denylist, and it catches #660's exact failure at the only moment anyone can still act. Good idea, wrong PR: it changes `read-continue-here`'s cleanup block, which would widen this diff into a third skill's behaviour. Filing it instead.

Issue #493, opened 2026-08-05
harness: shared-scratchpad filename collision + a non-aborting `cd` sent a review subagent's fixture into the real repo


pjt222:
## What happened

While adversarially reviewing #491, a review subagent built its test fixture **in
the working repository** instead of under `/tmp`, then ran `git add -A` and
`git commit`. The result landed on top of the PR branch:

```
11b90d75 de translation
 i18n/de/skills/demo-skill/SKILL.md | 15 +++++++++++++++
```

`demo-skill` is the fixture id from `scripts/test/normalize-i18n-fences.test.js`.
The file is a synthetic German "translation" of a skill that does not exist:

```
name: demo-skill
description: Eine Demo-Fertigkeit.
source_commit:            # empty
...
Uncommitted human prose.
```

The workflow prompt said, verbatim:

> IMPORTANT: never run `node scripts/normalize-i18n-fences.js --write` against
> the repo itself. Build a throwaway fixture repo under `/tmp` if you need to
> exercise write behaviour - `scripts/test/normalize-i18n-fences.test.js` shows
> exactly how.

The agent followed the *shape* of that instruction and ignored its location.

## Why this is the same bug as #486

#486's own "Contributing factors" section names this exact mechanism:

> The instruction not to modify files was prose in a prompt, with nothing
> enforcing it — the same failure class as #472 itself, where a rule stated four
> times was violated 1,307 times because nothing checked.

So the harness built to verify the fix for #486 reproduced #486's root cause. The
prompt was more specific this time (it named the directory, the tool, and the
file to copy) and specificity did not help, which is the finding.

## Blast radius, and what caught it

Contained: one file, one commit, never pushed. But it was on the local branch
HEAD for roughly ten minutes and would have been included in the next push.

It was **not** caught by `git status` — the agent committed, so the tree read
clean. Every direct check for stray writes passed. What caught it was
`npm test`, via `check-readmes`: the README's translation table counts
`i18n/<locale>/skills/*/SKILL.md` by existence, so the extra directory moved
`de` from 366/369 to 367/369 and the generated-file gate went stale.

A generated file whose content is derived from the corpus is an integrity check
on the corpus. That is worth knowing on purpose rather than by accident.

## Why `git status` was the wrong instrument

#486's lesson was "check `git status` before believing a number that moved". This
incident is the complement: an agent that *commits* defeats that check. The
durable question is not "is the tree dirty" but "is the tree what I left it as",
which means comparing HEAD too.

[...]

PR #859, opened 2026-09-16
chore: scrub a committed session UUID, and require absolute paths in agent destructive commands

pjt222:
Two safety commits, found by two different routes on the same day.

## Premises

| This is correct only if… | Verified where | Result |
|---|---|---|
| the session path is really in a tracked file | `git ls-files --error-unmatch` on the fixture | Holds — tracked, and this repo publishes to npm |
| scrubbing it does not change the tool's rendered output | `node tools/review-findings.mjs --verify` | Holds — 39/39; the paths sit in a `filesRead` array the renderer never emits |
| no other tracked file carries the shape | `git grep` for the session-path form and the UUID | Holds — 1 file before, 0 after |
| the relative-`rm` exposure is real rather than theoretical | audited 28 subagent transcripts | Holds — 55 relative calls, one of which targets this repo's live handoff by name |
| the template still parses after the edit | `wrapWorkflow` + `node --input-type=module --check` | Holds. Plain `node --check` fails on the file's top-level `return` and is the wrong ruler |

## 1 — `chore(tools)`: a committed session UUID

`tools/fixtures/review-r2-input.json:7` carried `/tmp/claude-1000/-mnt-d-dev-p-agent-almanac/<uuid>/scratchpad/` twice on one line.

**Severity low, stated plainly rather than inflated:** uid 1000 is the universal default and `/mnt/d/dev/p/` is already in this repo's own `CLAUDE.md`. The session UUID is the only new information and is an opaque identifier. What earns the commit is that this is precisely the class the four redaction skills exist to catch, sitting in the fixtures directory of the tool that renders review output.

Replaced with `/tmp/claude-<uid>/-<project-slug>/<session-id>/`.

**Found by the candidate `tools/enforce-redaction-gate.sh`** under development for #853 — the gate caught a real leak in this repository before shipping, which is the most useful thing it has done. It also unblocks that gate's `project-store-slug` shape, which could not be widened to reach the `/tmp/claude-<uid>/` form while a tracked corpus file matched it.

## 2 — `docs(safety)`: absolute paths in destructive commands

`REPO_SAFETY` required `mktemp -d` and a fatal `cd`. Both right; together they left `cd "$DIR" || exit 1` carrying all of containment, because a relative `rm` after a failed `cd` resolves against the repository — #493's mechanism from the deleting side instead of the writing side.

Audited before writing the rule. 28 subagent transcripts, 122 `rm` command lines:

```
risky absolute path (repo / $HOME / / / bare dot)        0
relative (rm -rf t, rm -f err.tmp, rm -rf nobin)        55
```

Every one safe, all for the same single reason. None of those names collides with a repo entry — **except one**: a reviewer exercising `read-continue-here`'s cleanup block ran `rm -f CONTINUE_HERE.md docs/CONTINUE_HERE.md`, relative, and `docs/` exists here. Had its `cd` failed, that takes the live 163 KB handoff. Its transcript shows the guard immediately above it, so it held and nothing was lost.

Nothing went wrong. One control was carrying all of it, and an absolute path costs nothing.

> `rm -rf "$DIR/fixtures"`, never `rm -rf fixtures`.

Both pointer sites get the canonical text rather than a summary — a preamble quoted in three places is three places to drift. `CLAUDE.md` also now separates two senses it had begun to conflate: an absolute path for **deleting** (this rule) and one for **writing** (the existing write-location sentence).

## Verification

```
npm run test:scripts            876/876, 0 fail
npm run validate:integrity      PASSED (7 pre-existing glyph/palette warnings)
npm run check-readmes           All files are up to date.
npm run check:tools-registry -- --verify   16 rows, 15 self-tests run
node tools/review-findings.mjs --verify    39/39
wrapWorkflow + node --input-type=module --check workflows/_template.mjs   OK
```

## Weakest claim

**That the rule will be followed.** It is a preamble, and the audit it rests on found agents complying with the *existing* preamble while still writing relative `rm`s — because the existing text never asked for absolute paths. A prompt is not a mechanism; `guard:snapshot`/`verify` remains the control that works when instruction fails. What this change buys is that the instruction now names the thing that actually matters, so an agent following it is protected by two controls rather than one.

Also unmeasured: whether any agent type ignores the preamble entirely. Every transcript audited carried it.

## i18n — in scope after all, and handled

The original body said *"No i18n — `guides/` has no mirrors and the two skills involved were not touched."* Both halves were wrong, and round 1 caught it.

`guides/` has 20 mirrors (5 guides × 4 locales); neither guide touched here is among those five, so the conclusion held by accident rather than for the stated reason. And `skills/create-workflow/SKILL.md` **is** in this diff and has four mirrors, all of which went stale the moment its English body moved. `check-translation-freshness.js` does report them — it runs inside the required `skills` job — but it runs `--warn` there, so its four rows cannot redden anything and they land among 2581 pre-existing STALE lines. The `translations` job that could go red is `paths: ['i18n/**']`-filtered and does not run on this PR at all. Detectable in principle, invisible in practice.

All four are byte-copy stubs, so the repair is mechanical and lands as the last two commits before merge: `npm run refresh:stubs -- skills create-workflow`, then the same command with `--stamp <sha>`. Verification on the merge head is `refresh:stubs --verify` printing `4 mirror(s), 4 stub(s), 0 diverged`.

One warning for whoever reads the checks afterwards: when `translations` runs here for the first time, the caveman backlog it reports is pre-existing and is not this PR's.

## Not in scope

No change to the candidate redaction gate, which is still in development and not shipping (see #853).

## Adversarial round 1 — `GATE: BLOCKING — 13 findings`

Reviewed by `advocatus-diaboli` on Opus, against a stamped bundle rather than the working tree. Two findings changed the substance of the PR:

**The rule as first shipped was unsafe.** `rm -rf "$DIR/fixtures"` with `DIR` unset expands to `rm -rf /fixtures`, and the preamble's own `cd "$DIR" || exit 1` does not catch it because `cd ""` returns 0 without moving. The old text bounded the worst case at the repository; the fix as first written unbounded it to `/`. Corrected to `rm -rf "${DIR:?}/fixtures"`, which refuses on unset *and* empty — measured identical on bash 5.2.21 and zsh 5.9. Inside the JS template literal it must be written `\${DIR:?}`, since `${` is interpolation there; the rendered string was extracted and checked (1518 chars, no residual escapes, one `${`), and the file still parses in the Workflow dialect.

**The audit figure was re-derived** — and the first re-derivation was wrong. See the round 2 section below; the claim that the original figures do not reproduce is **retracted**. Numbers to quote are in `tests/results/2026-09-17-repo-safety-rm-audit/RESULT.md`, not in this paragraph.

What survives unchanged is what the rule rests on: zero risky absolute paths in both measurements, and a real population of relative `rm` calls safe only because the `cd` held.

Also applied: the enumeration's "wrong by two" corrected to "drifted by one" (the second gap was opened by this PR itself); `guides/agent-best-practices.md` item 4 no longer contradicts itself, narrowed to the template's own assertion scope; the preamble's fifth rule added to the two sites that omitted it; and the near-miss re-described as run *while* exercising the cleanup block rather than by it, since the shipped block is already absolute and guarded.

Deferred to its own issue: `workflows/review-changes.mjs` spawns five shell-capable agents and carries neither the preamble nor a write-location line — a survivor of exactly this propagation class, and one a `REPO_SAFETY` grep structurally cannot find. Filed as #861, widened on measurement: all three shipped workflows, and `batch-generate-waves.mjs` spawns `general-purpose`, whose tool set is `*`.

## Adversarial round 2 — `GATE: BLOCKING`, and the blocking findings were against my own evidence

Round 2's blockers were not about the rule. They were about the artifact round 1 asked me to write, and they were right.

**The re-derivation read 106 of 739 transcripts — 14%.** `readdirSync(dir)` is not recursive, and under this project the transcripts sit at two depths: 106 directly under `<session>/subagents/`, and **633** under `<session>/subagents/workflows/wf_<id>/`. Every one of the 633 skipped files belonged to a **workflow-spawned** agent, so an audit written to justify the workflow template's own safety preamble excluded exactly the population that preamble governs. The cross-check probe had the same blind spot, so its four "independent counting rules" all counted the same 14%.

The transferable part is not "recurse". The probe *did* print a denominator, and 106 was an honest count of the files it opened. **A printed denominator proves what a scan read, never what it should have read.** The probe now prints a second number beside it — an independent recursive count of candidates on disk — and refuses when `scanned + filtered ≠ candidates`. Its bucket-sum self-check was also the same expression printed twice; strict matches are now counted separately and compared.

**Two claims are retracted.** First, "the published figure does not reproduce". It does: on the original claim's own terms (2026-09-16 alone) this measures 21 transcripts, 73 strict invocations and 41 relative, against a published 28 / 122 / 55 — same order of magnitude on every row, over a corpus that has since rotated. The all-time maximum across four counting rules is 145, not 39.

Second, and worse, **"zero naming a risky absolute path"**. Over the full corpus there are three, and one is unambiguous: a bare `rm -rf *`, risky whatever the working directory is, in a transcript dated 2026-09-16 — *inside the original audit's own window*. The one number the original got wrong was its zero, and this branch repeated it off a 14% sample. The other two rows are a workflow deleting its own worktree scratch directory, where the implemented rule (anything under the repository) is stricter than the documented one (the repository root); that gap is now stated rather than paraphrased.

**The corrected numbers strengthen the rule**: 48 relative calls rather than 7, 45 of them real deletes, and a real `rm -rf *`.

Also applied from round 2: the cwd assertion was still unbraced while the `rm` site had been braced — outside any repository `git rev-parse` prints nothing, so the unbraced form compares `""` to `""` and **passes**; braced at all five sites. `skills/create-workflow/SKILL.md` item 4 still carried the widened "before any destructive step" framing that round 1 narrowed out of both guides, and a `git grep 'before anything destructive'` could not see it because the survivor used different words. The `:?` prose is now qualified: it aborts the enclosing shell at top level, only the subshell inside `( )` or `$( )`, and it checks non-emptiness rather than absoluteness.

RESULT.md also gained a section on **what the classifier cannot measure**: it cannot distinguish `rm -rf "$DIR/x"` from `rm -rf "${DIR:?}/x"`, so it cannot score compliance with the rule it supports, and that bucket is 69 of 124 rows.

## Adversarial round 3 — `GATE: BLOCKING`, and the guard could not catch its own bug

Round 2 found that the re-derivation read 14% of its corpus. Round 3 found that **the guard added in response could not have caught it.**

`candidateCount` and the scan both came from the same `walk()`, so any blindness in `walk` was invisible to the comparison — while the file described it as "an independent recursive count". The reviewer reintroduced version 1's exact defect (one site, `if (isDir) walk(...)` → `continue`), left every new guard intact, and got `106 / 106 / 0 risky-absolute` at **exit 0**. The reassuring `(recursive, before any date filter)` in that output is a string literal asserting a property of the code, not a measurement of it.

The rule that survives three rounds, and the one worth taking away from this PR:

> **A self-check must compare two values produced by different code.** One value printed twice is not a check — at the level of a variable, or of a function.

Candidates are now counted with `readdirSync(root, { recursive: true })` — Node's own implementation, sharing no code with `walk()`, agreeing with `find(1)` at 739 — and the run refuses when the two disagree. Proven rather than asserted, on a two-file fixture whose nested file holds the risky row:

```
BASELINE  candidates 2, scanned 2, INVOKING 2, risky-absolute 1        exit 0
MUTANT    REFUSED: two independent traversals disagree — walk() found 1,
          node readdirSync({recursive:true}) found 2. One of them is blind.   exit 2
```

The mutant is version 1's bug, and the fixture reproduces its signature: it would have reported **zero** risky rows.

Three further blind spots the reviewer measured — each returning a clean plausible result while missing a file — are now refusals: a permission-denied subdirectory swallowed by a bare `catch`; a mistyped root beside a good one; a symlinked session directory skipped because `Dirent.isDirectory()` has lstat semantics.

**The prose no longer rests on a headcount.** Every published number has been restated twice in two days, and the reviewer measured this PR's own reviewer transcript at **31 of 130 strict rows — 24%** — because resuming an agent appends to one transcript and each round accumulates there. A fifth correction was likely. So the four sites now give the *mechanism* as the reason (a two-arm demonstration unchanged since round 1), state in one clause that a real population of relative `rm` calls exists, and point at RESULT.md, which owns the counts. A future correction changes one file instead of four.

Also applied: `cd "$DIR" || exit 1` was the last unbraced `$DIR` of three — in the very block whose bullet explains that `cd ""` returns 0 without moving — and is now braced everywhere, with a render check asserting zero lines teach the unbraced form. "45 of the 48" silently promoted two no-op invocations into real deletes and is now published as the three-way split: 3 splitting artefacts, 2 no-ops, **43** ordinary relative deletes. The near-miss section, lost in the round-2 rewrite while `CLAUDE.md` still cited it, is restored.

## The mirrors, and the check that sees them

The four `create-workflow` mirrors are refreshed and stamped in the last two commits. They were deliberately held back for three rounds because a stub refreshed against a body that then changes is stranded, and English moved in every round.

**The refresh is also the fence propagation**, and this is the part worth a reviewer's attention: `check-i18n-fence-parity.js` reported `OK` throughout, because it accepts a fence body from *any* historical English revision and therefore cannot see whether an edit reached the mirrors. The check that could see it is not in CI:

```
$ npm run check:fence-propagation -- --id create-workflow
# before:  4 of 4 alignable, 6 frozen of 6 — LAGS on fence #5 (`bash`) in all four
# after:   OK — 6 frozen fence(s) byte-identical across 4 mirror(s)
```

Anyone editing a frozen fence in English on this repository should run that command before merge; the gate that runs in CI is the one that cannot see the failure.

Verified on the merge head: `refresh:stubs --verify` → `4 mirror(s), 4 stub(s), 0 written, 0 refused, 0 diverged, 0 without fence_basis_commit`. Cap scan: English 328 lines, mirrors 334, ceiling 494.


pjt222:
## Path-scope declaration — lead session, 2026-09-17

Two sessions are active in `/mnt/d/dev/p/agent-almanac` today. Declared before this session's
first edit, per `coordinate-peer-sessions` Step 3. One-sided until the support session
acknowledges it; narrow rather than widen if it disagrees.

**Branches are the first constraint, not the paths.** One worktree holds one branch. The root
checkout is on `chore/safety-scrub-and-absolute-paths` (PR #859) and stays there — the support
session must not `git checkout` here. `main` is held by `.claude/worktrees/probe+722-truncation-notice`,
so the support session branches from `origin/main` **in its own worktree** under
`.claude/worktrees/<name>`, which needs three symlinks (`node_modules`, `viz/node_modules`,
`viz/renv/library`) and a 600 s timeout to create.

```text
Lead session (this one): PR #859 — chore/safety-scrub-and-absolute-paths in the ROOT checkout
  workflows/_template.mjs
  guides/agent-best-practices.md
  guides/creating-workflows.md
  skills/create-workflow/SKILL.md + i18n/{de,es,ja,zh-CN}/skills/create-workflow/SKILL.md
  tools/fixtures/
  CLAUDE.md  (shared — already edited on this branch; see below)

Support session: #858, #852, #855, #857 — its OWN worktree, branched from origin/main
  skills/enforce-redaction-gate/   (#858)
  skills/read-continue-here/       (#857)
  issue bodies for #852 and #855   (no tree edits)

Shared, ask before editing: CLAUDE.md, package.json, the four _registry.yml files
Nobody runs: git stash, git checkout -- <path>, git reset --hard, git add -A/--all/.
Nobody runs in the OTHER session's checkout: guard:release, guard:rebaseline
```

**The guard**: this session arms it only around a fan-out (the #859 review round) and releases
its own slot afterwards. If the support session sees a snapshot it did not arm, it reads a
`guard:verify` as reporting tree movement, never run liveness, and does not act on the
`git reset --mixed <baseline>` line that a red verify prints — that advice is written for
whoever armed it.

Staging is by explicit path in both sessions.

PR #871, opened 2026-09-18
feat(skills): consult-a-decision-oracle — gating an external probabilistic classifier

pjt222:
## What

One new skill in the `general` domain — `consult-a-decision-oracle` — on putting an external probabilistic classifier in front of a decision path without letting it take the decision over. Plus a `references/EXAMPLES.md`, a registry entry, regenerated READMEs, and (from earlier on the branch) a `.gitignore` rule for a local `.env` with a tracked template.

No new domain. `general` goes 25 → 26, `total_skills` 372 → 373, domain count unchanged at 67 — so this adds nothing to the viz-wiring backlog that `agent-commerce` is still sitting in.

## Why this shape

The starting request was to integrate TypeSafe AI's System One model (Jev) — a paid typed-decision HTTP API that returns calibrated scores. The obvious deliverable would have been a skill teaching how to call it. That deliverable is worthless, for reasons that are checkable rather than aesthetic:

**The vendor already ships that skill, and its own troubleshooting page argues against a downstream copy.** Under *"The agent invents request or response fields"*, `docs.typesafe.ai/agent-skill.md` says: *"A stale skill can cause this."* The vendor names a stale downstream skill as a cause of hallucinated fields.

**The vendor's own corpus drifts against itself.** Research for this change found four contradictions in one afternoon, each between pages the vendor itself links:

| Claim | Page A | Page B |
|---|---|---|
| Request token budget | `primitives.md`: one ~32,000 budget | `models.md`: 64k total, separate 32k |
| Batching speedup | `primitives.md`: 11.5× / 9.6× | linked cookbook: 12.2× / 10.0× |
| What the response `model` field carries | `api.md` examples: the alias | `models.md`: the versioned id |
| High-stakes confidence gate | `confidence.md`: `> 0.9` | `confidence-routing.md`: `> 0.85` (same action) |

A static copy does not merely inherit that drift; it freezes whichever side it read, across the English file and four translation mirrors whose code fences are byte-identical by construction, behind a gate that cannot detect semantic staleness.

**So the skill contains no vendor constants at all** — no field names, status codes, option caps, token budgets, or recommended thresholds. It contains method, which does not expire, and measurements that carry the model id and date they were taken from.

## What it actually teaches

All of it came from grading a real shipped integration, not from documentation:

- **Consult on one layer, gated on a precondition that is a correctness property.** The source integration changes 2 answers out of 87; its heuristic path is 43/43 on the cases it attests. Framing an oracle as a *replacement* would teach the opposite of what that code does.
- **Find the externally-graded rows you are already logging.** In the source case the corpus was discovered, not built: a log written to feed a circuit breaker had been accumulating externally-graded rows for 39 days before anyone noticed (verified against its commit history). Stated beside its inverse, because that one is never checked: *the signal you need may already exist by accident; the signal you built on purpose may already be dead.*
- **Report the majority-class baseline and per-class row counts beside accuracy.** On a skewed corpus the baseline is most of the apparent skill.
- **Derive the threshold from measured separation, report the gap's width, and write the reason as a calculation.**
- **Fail open at the call site, and check the failure is not silent.** A mitigation that swaps one silent failure for another is not a mitigation: a floating model alias silently invalidates a measured threshold; a pinned version silently dies when retired. Pin *plus* a tripwire.
- **Guard every value read from the environment, not only the credential.**

The degradation ladder is cross-referenced to `circuit-breaker-pattern` rather than restated — that skill already owns it.

## The weakest part, named rather than hidden

**Nearly all empirical content traces to one integration at one organisation.** The method generalises in principle; the evidence is one case. Two mitigations: every figure carries its model id and date at the point of use, and the grader taxonomy in `EXAMPLES.md` marks its single real case `[shipped]` and all thirteen other entries `[inference]`, per entry rather than in a preamble.

Second: the fixture as first committed in `1c8cdc4b7` shipped an interval that was **arithmetically impossible** — dropping the only incorrect row empties the wrong set, so no upper bound exists. The published numbers had been computed from a different split than the one the procedure defines. Fixed in `98141db13`, and **left described in the file rather than silently repaired**: a skill whose central instruction is *"write the reason as a calculation, then do the calculation"* published an interval its author had not calculated. It now sits next to a second instance of the same failure — a threshold documented for two months as the midpoint of `(0.54, 1.00]`, where the midpoint is `0.77`. A worked example of the failure beats an invented one.

## Validation

Every step in the skill's Validation section runs **offline with no API key**, because a reader without one must be able to run it. The fixture's default variant deliberately does **not** separate, so the assertion is that the procedure reaches its refusal branch — a fixture the procedure always passes is not a test of the procedure.

## Gates

| Check | Result |
|---|---|
| `validate:integrity` | PASSED |
| `validate:untagged-fences` | 0 untagged across 562 files |
| `validate:security` | 0 findings |
| `validate:line-endings` | OK |
| `audit-skill-sections` | no missing sections |
| `test:scripts` | 889 pass, 0 fail |
| `test:cli` | 139 pass, 0 fail |
| line count | 439 / 494 ceiling |
| `check-readmes` | up to date (regenerated, not hand-edited) |

## Follow-ups (not this PR)

- Translation scaffolds for `de`, `zh-CN`, `ja`, `es` — after merge, per the repo rule that a scaffold is a byte-copy taken once the English is final.


pjt222:
Adversarial round complete; all findings addressed in `d1ae77bea`. Its central finding was that the first fix moved the defect rather than closing it — selecting override rows was right, grading them by whether the *oracle* was right was not, because a both-wrong override is outcome-invariant and destroys a real gap. Step 4 now counts the gate's harm rather than the oracle's errors.

Also from that round: the interval's closed endpoint was not free under the only operator either file exhibited; width without `n` repeated the error width was added to fix; a discovered corpus can be censored by the live gate's own threshold; three `On failure:` branches were not the negation of their `Expected`; "calibrated" was wrong and unnecessary; and the one command the file told you to run could not run, because `fixture.json` was fenced rather than shipped. It now ships alongside `separation.py`, whose three regression arms were mutation-tested — reverting the predicate kills arm 3 and exits 1 while arms 1 and 2 stay green.

One defect found by a red check rather than by review: `readmes` went stale on `SECURITY.md` while the same command exited 0 locally, because `generate-readmes.js` walks the filesystem rather than the index and had counted a gitignored `.pyc` into the published security surface. Filed as #872 with a clean repro — a gitignored executable is not merely counted but *named* as a shipped script.

**Scope note for the record:** the adversarial round graded `98141db13`. The three commits after it address that round's findings and have not themselves been through a fresh adversarial pass. All 15 contexts are green on `9ac9e5d4b`, and the regression arms plus the gates are the standing evidence for those commits.

Issue #868, opened 2026-09-17
[LOW] check-tools-registry walks the filesystem rather than asking git, so a gitignored build artefact reddens a required gate


pjt222:
## Summary

`scripts/check-tools-registry.js` enumerates `tools/` from the filesystem and refuses any entry that is not a plain file. `tools/__pycache__` is **gitignored** and is still enough to fail the check — and `check:tools-registry` runs inside `validate:integrity`, which is one of the five required contexts.

This repository already has the lesson written down for a different checker, in `CLAUDE.md` § Viz Deploy Model: *"walking the filesystem instead of asking git scans 7,177 files instead of 72 and reaches the annotated examples vendored into `viz/renv/library/`."* Same shape, different gate.

## Reproduction

Measured 2026-09-17 on `feat/853-redaction-port`:

```
$ python3 -c "
import importlib.util as u
s=u.spec_from_file_location('ra','tools/redact-artifact.py'); m=u.module_from_spec(s); s.loader.exec_module(m)"
$ ls tools/__pycache__/
redact-artifact.cpython-312.pyc
$ git check-ignore -q tools/__pycache__ && echo ignored
ignored
$ npm run check:tools-registry
FAIL: not a plain file under tools/: tools/__pycache__ -- a tool is one flat file; a subdirectory
  or symlink there is representable by no row (NOT_TOOLS is the exact exemption set)
FAIL: 18 row(s) (18 active) against 18 plain file(s) under tools/, three directions (1 not a plain file)
```

## When this fires

Not on ordinary use — measured, and the boundary is sharp:

| how the Python tool is used | `tools/__pycache__` written? |
|---|---|
| run as a script (`python3 tools/redact-artifact.py …`) | **no** |
| **imported by path** from another process | **yes** |
| imported with `PYTHONDONTWRITEBYTECODE=1` | no |

So it fires whenever anyone *imports* a Python tool rather than running it — a reviewer probing it, a test harness, or a consumer reusing `tools/redaction-lib.py`, which is a library and is meant to be imported. A module cannot defend itself here: Python decides whether to cache a file before that file's body runs, so an in-module `sys.dont_write_bytecode = True` protects everything it imports and never itself.

## Why it is LOW and not higher

Nothing is lost and no leak follows; the gate fails closed and the fix is `rm -rf tools/__pycache__`. It is filed because a required gate that a *gitignored* artefact can redden will keep costing minutes to diagnose, and because the same class is already documented for another checker.

[...]

Issue #830, opened 2026-09-15
[LOW] tools-registry: a gitignored build-residue directory in tools/ reds the required integrity check


pjt222:
## Summary

`check:tools-registry` enumerates `tools/` with `readdirSync` and fails on anything that is not a plain file. `__pycache__/` is gitignored (`.gitignore:61`) but is not in the gate's exemption set, so importing any Python tool as a module creates a directory that reds `integrity` — a **required** status check — while `git status` reads clean.

Reproduced accidentally on 2026-09-15 while building `tools/append-only.py` (#829): a probe imported the tool with `importlib`, and the next `npm run check:tools-registry` printed

```
FAIL: not a plain file under tools/: tools/__pycache__ -- a tool is one flat file; a subdirectory or symlink there is representable by no row (NOT_TOOLS is the exact exemption set)
FAIL: 15 row(s) (15 active) against 15 plain file(s) under tools/, three directions (1 not a plain file)
```

`rm -rf tools/__pycache__` clears it. Nothing in the repository imports a tool today, so no current path triggers this — it is a landmine under the obvious next step, which is a `scripts/test/<id>.test.js` or a pytest that imports a tool rather than shelling out to its `--verify`.

## Findings

1. **[LOW]** `scripts/lib/tools-registry.js:190` — `readdirSync(join(root, TOOLS_DIR))` walks the filesystem, so a gitignored path is in scope for a gate whose other two directions are about tracked content.
2. **[LOW]** `scripts/lib/tools-registry.js:41` — `NOT_TOOLS` is a closed set of three literals (`README.md`, `fixtures`, `_registry.yml`). Closed is the right design (a prefix rule would silently absorb a new spelling), but it means every legitimate non-tool entry has to be named, and build residue was never named.
3. **[NOTE]** The failure is *correct behaviour reported against the wrong subject*: the gate is right that the directory is representable by no row. The question is whether a path git is told to ignore should reach the gate at all.

## Acceptance criteria

- [ ] Decide and record which of the two fixes applies: add `__pycache__` to `NOT_TOOLS`, or skip paths git ignores. Note the trade-off in the commit — the gate must still see an **untracked** new tool file, which is a real defect it catches, so "ask `git ls-files`" is not a drop-in substitute for the walk.
- [ ] Whichever is chosen, `python3 -c "import sys; sys.path.insert(0,'tools'); import capgeom"` followed by `npm run check:tools-registry` exits 0.
[...]

Issue #691, opened 2026-08-19
SECURITY.md generation: an unwatched input, untested exclusion, and unmarked static claims


pjt222:
## Summary

Four findings from the adversarial review of PR #686 that were correctly judged not-blocking, filed so they are not lost with the review.

## Findings

1. **[MEDIUM] `generateSecuritySurface` has an input no trigger, gate, or derived check can see.** It counts top-level files in `scripts/` with `readdirSync`. `update-readmes.yml`'s trigger paths cover `skills/**` (the Bash-count input) and the named generator modules, but nothing under `scripts/` generally; `validate-readmes.yml` is outputs-only by design; and `check:generator-inputs` derives from the **import graph**, which a `readdirSync` is not part of.

   Scenario: add `scripts/check-foo.js` — which happens without a `package.json` edit (`coverage-matrix.js`, `measure-tag-sequence-parity.js`, `normalize-content-style.js` and `bulk-scaffold-caveman.sh` have no npm entries, and the directory went 27 → 33 during the lifetime of #600). SECURITY.md is then stale with no healer at the causing commit, and it surfaces only when the daily schedule goes red with nothing to auto-repair it — violating `update-readmes.yml`'s own stated rule that drift is healed at the commit that caused it.

   Fix: add `scripts/*` (single star, matching the `readdirSync` non-recursive scope, and firing on deletions too) to the healer's paths.

2. **[MEDIUM] Static prose inside the AUTO markers borrows authority it has not earned.** Two of the claims in the generated section are now asserted at generation time — that `scripts/` is absent from `package.json`'s `files`, and that the three named mutating tools exist — and they throw if false. The rest is still static: the `viz/` bullet, the `.claude/` bullet, and the prose framing around the derived numbers.

   The hazard is specific: a reader cannot tell which half of an AUTO block is derived. A sentence that survives `check-readmes` reads as checked. Worth either deriving more of it or marking, in the file, which claims are asserted.

3. **[MEDIUM] The registry-not-directory property is celebrated and untested.** Enumerating `skills/_registry.yml` rather than walking `skills/*/SKILL.md` is what excludes `_template` and keeps the figure at 228 of 370 rather than 229 of 371 — the review confirmed the walk finds exactly 371. But `skillsDeclaringBash` lives in `generate-readmes.js`, which executes its whole pipeline on import, so nothing can test it. A future refactor to a directory walk ships green and the published figure silently gains one in each term.

[...]

PR #874, opened 2026-09-21
fix(scripts): enumerate content trees by asking git, not by walking disk

### LINKED ISSUE #872: generate-readmes walks disk, not git: a gitignored file is counted and named in SECURITY.md as shipped
## Summary

`scripts/generate-readmes.js` builds its content inventory by walking the **filesystem**, not the git index. A gitignored file under a content tree is therefore counted, and can be **named**, in `SECURITY.md`'s generated security-surface paragraph — under a sentence that asserts "All of it ships", which a gitignored file by definition does not.

Found while shipping #871: running a Python asset under `skills/` left a `__pycache__/`, and the committed `SECURITY.md` went out claiming **19** non-Markdown files including a `.pyc`. CI, checking out a clean tree, computed 18 and reported the file stale.

## Two defects, one cause

**1. The count and the extension list are wrong.** Measured on a clean tree at `9ac9e5d4b`:

```
$ npm run check-readmes                       # clean tree
EXIT=0

$ mkdir -p skills/consult-a-decision-oracle/references/__pycache__
$ printf 'x' > skills/consult-a-decision-oracle/references/__pycache__/repro.pyc
$ git check-ignore -q .../repro.pyc && echo ignored
ignored
$ git status --porcelain                      # tree still reads clean
$ npm run check-readmes
STALE: SECURITY.md
EXIT=1
```

Regenerating with the artifact present publishes:

```diff
-plus **18 files that are not** (.bib, .json, .py, .webp)
+plus **19 files that are not** (.bib, .json, .py, .pyc, .webp)
```

**2. An ignored executable is named as a shipped script.** Same setup with a `.py`:

```
$ printf '#!/usr/bin/env python3\nprint(1)\n' > .../__pycache__/evil.py
$ git check-ignore -q .../evil.py && echo "gitignored -> does NOT ship"
gitignored -> does NOT ship
$ node scripts/generate-readmes.js && rg -o 'executable scripts:[^.]*' SECURITY.md
executable scripts: `skills/consult-a-decision-oracle/references/__pycache__/evil…
```

So a file git deliberately excludes is published in a security document as a shipped executable. That is a false statement about the artifact, in the document whose job is describing the artifact accurately.

## Why the local check cannot catch it

`check-readmes` compares the generator's output against the committed file. With the artifact on disk, both sides read the same contaminated input, so the check agrees with itself and exits 0. Only CI — a clean checkout — disagrees. A maintainer sees a green local check and a red CI job with no visible cause, which is how this cost a debugging cycle rather than being obvious.

This is also the failure mode the repo's own rule anticipates: *"an unexplained stale generated file proves something about the corpus — investigate such staleness before regenerating it away."* Regenerating first produces a correct diff and silently discards the finding.

## Scope

Anything the generator walks under a content tree is affected, not just `__pycache__`. Plausible everyday instances: `.DS_Store`, editor swap files (`*.swp`), `*.bak`, `node_modules/` inside a skill asset dir, `.Rcheck/`, any local scratch output an author leaves beside a skill. All are in `.gitignore` and all would be counted; those with an executable extension would be named.

## Suggested fix

Have `nonDocumentationFiles` / the inventory walk in `scripts/lib/skills-inventory.js` enumerate from git rather than from disk — `git ls-files -z` over the content trees — so the inventory describes the committed artifact, which is what `SECURITY.md` claims to describe. Falling back to the disk walk outside a git checkout is fine; the published document is always generated inside one.

A cheaper variant, if the disk walk must stay: filter the found set through `git check-ignore --stdin` and drop ignored paths.

## Acceptance criteria

- [ ] With a gitignored artifact present under a content tree, `npm run update-readmes` produces byte-identical output to a clean tree
- [ ] A gitignored executable is never named in `SECURITY.md`'s executable-scripts list
- [ ] A regression test asserting both, with the artifact created and removed by the test
- [ ] The test is proven able to fail — confirm it goes red against the current disk-walking implementation before the fix lands

## Not in scope

The `SECURITY.md` content at `9ac9e5d4b` is already correct; #871 removed the artifact and restored the count. This issue is the generator, not the current file.

pjt222:
## Summary

Closes #872, #868 and #830 — three separately filed defects that are one class. Each of three enumerators built its answer by walking the filesystem, so a **gitignored** path reached a question about the committed artifact:

| Issue | Call site | What a gitignored path did |
|---|---|---|
| #868 / #830 | `scripts/lib/tools-registry.js` (`checkParity`) | `tools/__pycache__` reddened `integrity`, one of the five **required** contexts, while `git status` read clean |
| #872 | `scripts/generate-readmes.js` | a `.pyc` was **counted** in `SECURITY.md`, and a `.py` would have been **named** there as a shipped executable script |

#868 and #872 were produced by the same act one day apart — running a Python file with `importlib` rather than as a script. `CLAUDE.md` § Viz Deploy Model already carried the lesson in prose; three enumerators still walked disk. Prose did not transfer it, so this hoists the rule into a module.

> **This body describes the design at `f3e529840`.** Four earlier designs were replaced during review: `git ls-files --cached --others --exclude-standard` (replaced on a measurement), a single "tracked or not ignored" rule for every consumer (replaced on a measurement), a recursive enumerator no consumer called (deleted), and escaping a pathspec-metacharacter candidate rather than refusing it (reverted on a measurement). The comment thread has the history. Every figure below traces to a script under `tests/results/2026-09-21-git-files-enumeration/` or to a command quoted in place; three are marked HISTORICAL there because they measure an implementation this PR deleted.

## Two questions, so two rules

The rule was one rule until the review measured the premise under it. **"A gitignored file does not ship" is false.** With a `files` array and no `.npmignore`, npm packs the working tree:

```
$ npm pack --dry-run --json    # tree carrying the #872 artefact plus an untracked sibling
packed files: 621
pycache in pack: [".../__pycache__/evil.py", ".../__pycache__/repro.pyc"]
fresh-untracked in pack: [".../fresh-untracked.py"]
```

A *local* pack ships the ignored `.py`, the ignored `.pyc` and the untracked `.py` alike. The *release* pack contains none of the three, because CI checks out a commit and there the disk **is** the tracked set. So one rule describes neither artifact.

| Consumer | Question | Function |
|---|---|---|
| `checkParity` — the tools gate | what is in this directory that git is not ignoring | `topLevelEntries` |
| `nonDocumentationFiles` — SECURITY.md | what is in the index, which is what the next commit will contain | `listTracked` |
| the per-locale translation counts | either serves; uses the ignore rule | `topLevelEntries` |

The gate keeps the working-tree rule because #830's acceptance criteria require it: a **new tool file with no registry row is untracked** at the moment its author runs the gate, and an index-based listing would turn a defect the gate exists to report into one it cannot see. SECURITY.md takes the index rule, and the generated sentence now carries its own basis:

> All of it ships — **counted from the index, which is what the next commit will contain and what a release is packed from; a local `npm pack` packs the working tree instead and can include files this count excludes.**

Published count unchanged at 18.

## The ignore rule, and why `check-ignore` rather than `ls-files`

**The directory listing, minus what git ignores.** The disk is the candidate source; `git check-ignore` is the only filter. Nothing here decides what is ignorable — a hand-rolled matcher would be a fourth glob implementation in a repository that already documents how easy `*`-crosses-`/` is to get wrong.

`ls-files --cached --others --exclude-standard` was the first implementation and it is **190× slower** here, on the path of a required context. One historical run at `b8eee5b7e`, when both implementations existed side by side, over 46 sites, on this checkout (WSL2, `/mnt/d` is 9p/drvfs):

```
ls-files --cached --others --exclude-standard   32612 ms
readdirSync + one check-ignore batch per call     3299 ms
readdirSync alone, no ignore rule at all           173 ms
```

Not spawn cost — `git rev-parse` measures 36 ms. `--others` walks and stats the working tree and loads a 13,492-entry index on every call. The two implementations were run against each other before the replacement: identical answers on 46/46 call sites and 4/4 content trees. **That comparison needs both implementations, so no committed script reproduces it; `RESULT.md` marks it and the ext4 timings HISTORICAL with the sha they were taken at.**

Measured on git 2.43 rather than assumed, and re-derived independently by the review over nine ignore mechanisms: a **tracked** file matching an ignore pattern is not reported (so no tracked-set union is needed), a directory matching `build/` **is** reported when asked about as `build`, nested `.gitignore` / `!negations` / `.git/info/exclude` are all honoured, and exit 1 means "no path matched" while 128 means fatal.

## It refuses rather than answering without the rule

There is no disk fallback. An earlier revision fell back to the unfiltered listing whenever git could not be asked, and the review measured what that bought: a `git` failing the way `safe.directory` does (exit 128, "detected dubious ownership") put `tools/__pycache__` back into `notPlainFile` and a gitignored `evil.py` back into the inventory — silently, behind a `source: 'disk'` marker no consumer read. Its one named beneficiary did not exist either: `scripts/` is in `REPO_ONLY`, no shipped module imports this, and a package installed under a consumer's `node_modules/` sits *inside* their work tree where `node_modules` is ignored, so the fallback would have answered `[]` with certainty.

Every failure that is not git's documented "no path matched" throws, with git's own stderr in the message. Walk errors likewise: only `ENOENT`/`ENOTDIR` mean "nothing here"; `EACCES` throws, because git exits 0 with a mere `warning: could not open directory` and nothing downstream would refuse on its behalf.

**Four ways `check-ignore` answers something else**, all measured, one of them guarded:

- **W1 — a candidate is a pathspec, and no form of the question fixes it**, so such a name is **refused**. Sent RAW, `tools/x*y.log` comes back not ignored when a tracked sibling glob-matches the name. Sent ESCAPED, it comes back not ignored whenever the PATTERN carries the metacharacter, because `check-ignore` matches the pattern against the pathspec string as typed. Eight one-pattern fixtures with no tracked sibling anywhere: **RAW agrees with `git status --ignored` 8/8, ESCAPED 2/8**. `--literal-pathspecs` is no escape either — rejected outright, exit 128, for ordinary candidates too. The refusal names the file, and `\` is in its class: that was the member the first version missed, and an intermediate revision escaped instead of fixing it. The matrix is committed as `escaping-matrix.mjs`.
- **W2** — `core.ignorecase` is `true` on this mount and `false` on a Linux runner.
- **W3** — `core.excludesFile` and `.git/info/exclude` reach the answer, because git is answering.
- **W4** — a batch refuses as a unit (a path *through* a symlink is exit 128 for the whole call), which is why the thrown message carries stderr.

W2–W4 are recorded in the module header rather than guarded: they are properties of asking git at all, and a "works on my machine" should be diagnosed rather than investigated.

## Coverage, and what is honestly not covered

The three published counts live in `scripts/lib/tree-counts.js` taking `root`, and `tree-counts.test.js` asserts the **numbers** against a git fixture. That is a change of instrument, not a rename: the previous "wiring proof" was a scan of the generator's source, and the review measured it **green with the #872 defect fully restored** (second import line, one call site reverted, gitignored `scripts/local-probe.js` planted) and **red on a harmless specifier change**. A test that fails on a refactor and passes on the defect is a denylist.

Proven against that exact scenario, four arms, two of them controls, in a clone:

```
A  unmutated, clean                exit 0   pass 5
B  unmutated + the ignored file    exit 0   pass 5      <- does not flag any ignored file
C  mutated + the ignored file      exit 1   fail 2      <- the scenario the old scan passed
D  mutated, clean                  exit 1   fail 1      <- fixture alone is enough
```

**Not covered, stated rather than implied:** the one-line call site *inside* `generate-readmes.js` — `const scriptFiles = scriptFileCount(ROOT);` — is reviewed by eye, not tested. That file runs its pipeline and `process.exit` at import, so nothing can import it; the source scan that remains is labelled `TRIPWIRE (not the coverage claim)` in its own name, and the review demonstrated it can be bypassed with `opendirSync`. Making that function testable is #877.

**Sixteen mutants, every count re-taken at this sha under `npm run test:scripts`** — the command CI runs, which is what `CLAUDE.md` § Proving a Gate Can Fail requires. The plan is committed as `tests/results/2026-09-21-git-files-enumeration/mutation-plan.tsv`, so the table below is one command rather than a claim:

```
bash tools/mutation-envelope.sh --test 'npm run test:scripts' \
  --plan tests/results/2026-09-21-git-files-enumeration/mutation-plan.tsv
```

Each mutates a value so the module runs to completion, because a crash proves the code is reached, not that its effect is asserted.

| Mutation | File | Verdict |
|---|---|---|
| disable the `topLevelEntries` ignore skip | `lib/git-files.js` | KILLED by 9 |
| treat every git failure as "nothing ignored" | `lib/git-files.js` | KILLED by 2 |
| treat exit 128 as "nothing ignored" | `lib/git-files.js` | KILLED by 2 |
| add `EACCES` to the absent-means-empty set | `lib/git-files.js` | KILLED by 1 |
| disable the pathspec refusal entirely | `lib/git-files.js` | KILLED by 4 |
| drop `*` from the refused class | `lib/git-files.js` | KILLED by 1 |
| drop `?` from the refused class | `lib/git-files.js` | KILLED by 1 |
| drop `[` from the refused class | `lib/git-files.js` | KILLED by 1 |
| drop `\` from the refused class | `lib/git-files.js` | KILLED by 1 |
| add a space to the refused class (over-broad) | `lib/git-files.js` | KILLED by 1 |
| `listTracked` also asks `--others --exclude-standard` | `lib/git-files.js` | KILLED by 2 |
| `topLevelEntries(root, TOOLS_DIR)` → `'scripts'` | `lib/tools-registry.js` | KILLED by 4 |
| `listTracked(root, tree)` → `'i18n'` | `lib/skills-inventory.js` | KILLED by 8 |
| `scriptFileCount` → enumerate `workflows` | `lib/tree-counts.js` | KILLED by 2 |
| `workflowFileCount` → enumerate `scripts` | `lib/tree-counts.js` | KILLED by 2 |
| `localeTranslationCounts` → `localeRel` | `lib/tree-counts.js` | KILLED by 1 |

**Five rows are one per member of a denylist, plus one over-broad control**, because a denylist is only as strong as its weakest member and this one shipped incomplete: `\` was missing, and `?` and `[` were pinned by nothing at all. The over-broad row exists because the review measured the control's honest size — a guard widened to reject a *hyphen* dies to nine tests, since the real catalogue is full of hyphenated names, but one widened to reject a **space** survived 912/912, because no fixture and no enumerated name carried one. The control arm holds a space now, and that mutant dies.

Three things this table records that earlier revisions of it got wrong, each worth more than the rows:

**The `--allow-broad` waiver is gone, because it was an artefact of my test command.** Against a single test file the first mutant kills 7 of 10 (70%) and trips the tool's broadness heuristic; against `npm run test:scripts` it is a clean kill. A waiver that disappears when you measure correctly was never a waiver. The envelope now fixes the command once for the whole plan, so a table cannot mix them again.

**A mutant SURVIVED and found a vacuous arm of mine.** Sending a candidate unescaped changed no test result, because that arm's fixture had no *tracked* `.log` file — `git add -A` under a `*.log` rule adds none — and the index-side failure only fires when the candidate, read as a glob, matches something in the **index**. Git read each name literally, raw and escaped agreed, and the arm asserted the right answer for the wrong reason. Re-reading it would not have found that.

**And the design that arm was protecting turned out to be wrong anyway.** The measurement that exposed the vacuity is the one that reversed the design: escaping fixes the index-side case by breaking the pattern-side case, and is wrong in six of eight fixtures where raw is wrong in none — so the module refuses such a name instead, which is what its first revision did. That row is no longer in the table, because the code it mutated is not at this sha.

## Test fixtures

Nine fixtures across `skills-inventory.test.js` and `tools-registry.test.js` were bare `mkdtemp` trees exercising the disk fallback rather than the production path. They are git repositories now, through `scripts/test/_git-fixture.js`.

Two things that helper gets right because the review measured them wrong first:

- **`cleanEnv` covers the write side; `isolateGitEnv` covers the read side.** The module under test spawns git with `process.env`, so one line in a developer's `$XDG_CONFIG_HOME/git/ignore` matching a fixture filename reddens a suite with nothing in the diff to explain it. Each enumerating suite deletes every `GIT_*` key and moves `HOME`/`XDG_CONFIG_HOME` at module scope — per suite, because `node --test` isolates files by process, and production must keep inheriting.
- The `broken` fixture — the one tree not migrated — **passed for two reasons** afterwards: the CLI returns 2 for any throw out of `loadRegistry`, and "not a checkout" is now also a throw. It is a repository now and the arm asserts the captured message names the parse failure.

Three arms exist because the review asked for them by name: a symlinked directory (a path through one is exit 128 for the whole batch, and asking only about immediate children is what stops one ever being generated), a `git` that **exists and fails** via a PATH shim (the previous refusal arm used a bare directory, which answers a different question), and `listTracked` beside `topLevelEntries` on a fixture where the two sets genuinely differ.

One function was **deleted** rather than fixed: a recursive `listNonIgnored` that no consumer called. The gate is top-level by construction and the inventory asks the index, so the recursive-walk-plus-one-batch shape — which the module header spent a section justifying, with a "~430 spawns, ~30 s" cost argument — was code nothing ran, kept because its comment argued well for it. That is the same defect this body was corrected for one round earlier.

## Verification

```
npm test                       EXIT=0   (912 script tests, 141 CLI)
npm run validate:integrity     EXIT=0
npm run check-readmes          EXIT=0
npm run check:tools-registry   EXIT=0   with tools/__pycache__ present
npm run check:generator-inputs EXIT=0   (exit 1 before the two libs were listed)
npm run validate:line-endings  EXIT=0
```

Each run bare and read from `$?`, never through a pipe.

Both defects, gone, and both negative directions intact: with `tools/__pycache__` present `check:tools-registry` exits 0, and it still exits 1 for a non-ignored subdirectory and for an untracked tool file with no row; with a gitignored `.pyc` and `.py` under a skill's `references/`, `check-readmes` exits 0 and regeneration produces no diff, where both previously went stale.

Enumeration cost on this mount is 3.1–3.5 s against a ~170 ms `readdirSync` baseline — a ~3 s delta on a full generator run, stated rather than hidden. Over **42** sites, not the 46 the historical run used: the probe derived its locales from the DIRECTORIES under `i18n/` (which includes `glossaries`) where the generator iterates `_config.yml`'s `supported_locales`. The four extra were absent directories that spawn no git process, so the timings stood and only the count was wrong; the probe reads `_config.yml` now. It varies between runs — quote a run, not this sentence.

**On ext4 — a hosted runner — this design is SLOWER than the one it replaced, and saying so is the point.** The review measured `generate-readmes.js --check` at 0.33 s before any of this, 1.2 s under the first design, and 1.87/1.77/1.80 s under this one. Fewer git processes (31 against 98) and more wall time — the mechanism behind that is NOT measured, only the timings are. The drvfs number is 10x the other way, which is where the maintainer runs it and where a required gate was taking 32 s. Both numbers are the trade; neither is the headline on its own.

## One tool, promoted at its fourth typing

`tools/mutation-envelope.sh` runs a plan of mutants under one test command and fails when any row is not a clean kill. It exists because this PR's review rounds typed that loop four times as a throwaway heredoc, and two of the losses reached a published table: counts taken under a narrower command than the table claimed, and an `--allow-broad` waiver that was an artefact of the same mistake. Its self-test builds a repository with one COVERED and one UNCOVERED line, so the survivor path — the one that matters — is exercised rather than assumed; it caught two defects in the tool before the registry row was written. `verify_in_ci: false`, with the reason recorded: a full mutation cycle takes minutes.

## Follow-ups filed, not fixed here

- **#876** — a local `npm publish` from a tree carrying a `__pycache__` ships it, and `assertInterpretable` refuses the glob negation that would stop it. Found by the same measurement that produced the two-rule split.
- **#877** — `generate-readmes.js` cannot be imported, so its call sites cannot be tested.

`scripts/check-generated-artifacts.js` is where this module's shape came from — its `git()` / refuse-on-failure pattern was lifted rather than reinvented. Adopting the shared module there is a follow-up, not this PR.


pjt222:
## Redesign after a measurement: `check-ignore`, not `ls-files --others`

The PR body above describes the first implementation. It was **190× slower than the `readdirSync` it replaced**, on the path of a required context, and my own weakest-claim list is what turned that up. Measured on this checkout (WSL2, `/mnt/d` is 9p/drvfs), over the 46 call sites `generate-readmes.js` makes:

```
ls-files --cached --others --exclude-standard   32612 ms
readdirSync + one check-ignore batch per call     3291 ms
readdirSync alone, no ignore rule at all           173 ms
```

The cost is not process spawn — `git rev-parse` measures 36 ms. It is `--others`, which walks and stats the working tree and loads a 13,492-entry index on **every** call. `check-ignore` consults the ignore rules and the index and stats nothing.

**The rule is now "the directory listing, minus what git ignores".** The disk is the candidate source; `git check-ignore` is the only filter. That still excludes the gitignored artefact and still sees an untracked new tool file, which is what #830's acceptance criteria protect. Both implementations were run against each other before the replacement landed: **identical answers on 46/46 call sites and 4/4 content trees**.

`missing` and `presentOnDisk` are gone with `--cached` — the disk is the candidate source, so "in the index but not on disk" is out of scope, which is exactly what the walk being replaced did. `checkParity` still reports a registry row whose file was deleted locally, through its own `existsSync`, and there is a test arm for it.

### Behaviour measured, not assumed (git 2.43)

The probe is quoted in the module header; these four facts are what the design leans on.

| Question | Measured |
|---|---|
| Is a TRACKED file matching an ignore pattern reported as ignored? | **No** (`--no-index` reports it; the default consults the index) — so no tracked-set union is needed |
| Is a directory matching `build/` reported when asked about as `build`, no trailing slash? | **Yes** — which is how `topLevelEntries` asks |
| Nested `.gitignore`, `!negations`, `.git/info/exclude`? | All honoured, because git answers |
| Exit codes | 1 = no path matched, 128 = fatal — told apart by status, never by matching a message |

### One batch per call, never a spawn per directory

Asking per directory would be ~430 spawns for the four content trees, about 30 s at the ~70 ms a spawn costs here. Batching is correct as well as faster: git reports a path *inside* an ignored directory as ignored, so pruning the walk is an optimisation rather than a requirement.

### Re-verification after the redesign

```
npm run validate:integrity        EXIT=0
npm run test:scripts              EXIT=0   (899 tests)
npm run check-readmes             EXIT=0
npm run check:tools-registry      EXIT=0   with tools/__pycache__ present
npm run check-readmes             EXIT=0   with a gitignored .pyc and .py under a skill's references/
```

Enumeration cost over the same 46 sites is now **3729 ms** against the 172 ms `readdirSync` baseline — a 3.5 s delta on a full generator run, stated rather than hidden.

### Mutants re-run against the new design

| Mutation | File | Verdict |
|---|---|---|
| `found.filter(!ignored.has(rel))` → `filter(Boolean)` | `lib/git-files.js` | KILLED by 4 — **`--allow-broad` waived**, see below |
| delete the `ignored.has(...)` skip in `topLevelEntries` | `lib/git-files.js` | KILLED by 2 |
| `topLevelEntries(root, TOOLS_DIR)` → `'scripts'` | `lib/tools-registry.js` | KILLED by 4 |
| `listNonIgnored(root, tree)` → `'i18n'` | `lib/skills-inventory.js` | KILLED by 8 |
| three `topLevelEntries(ROOT, …)` sites, one each | `generate-readmes.js` | KILLED by 1 each (file unchanged by the redesign) |
| re-add `readdirSync` to the `fs` import | `generate-readmes.js` | KILLED by 1 |

**The waiver is named rather than buried:** the first mutant was reported `SUSPECT KILL` at 4 of 10 tests (40%, "broad for one line"). It is not a crash — the mutant parses and runs to completion, and the four arms that die are the four that assert the ignore rule (the accept rule, the tracked-but-ignored file, the nested-`.gitignore` trio, and the skills-inventory wiring). `--allow-broad` was passed deliberately, with that reading; the tool's own guidance is that both SUSPECT signals are waivable and that SUSPECT names a doubt for a human rather than a verdict.

pjt222:
## The npm-ships predicate is unchanged — measured against the walk it replaced

My own weakest claim 3 was the ancestor-prefix loop in `shippedFilesUnder`: the recursive walk tested each directory before descending, and a flat listing has to test those prefixes explicitly or a directory negation stops excluding anything. An off-by-one there changes what `SECURITY.md` says ships, and nothing else would notice — the published number is whatever the code says.

So it is compared rather than argued. The original `walk` was lifted verbatim out of `7f3ffcf5d^` and run beside the new rule over the real corpus, with negation sets chosen to sit on the boundaries:

```
ok   the real package negations
ok   a file negation at depth 1            (agents/_template.md)
ok   a directory negation at depth 1       (skills/_template/)
ok   a directory negation at depth 2       (skills/create-skill/references/)
ok   a file negation at depth 3            (.../references/separation.py)
ok   a negation naming the TREE itself     (skills/)
ok   a prefix that matches no boundary     (skills/create-sk)
ok   no negations at all

nonDocumentationFiles: 18 paths; old walk on the same tree: 18
identical: true

cases compared: 8 x 4 trees; disagreements: 0
```

The `no negations at all` and `a prefix that matches no boundary` rows are the controls: without them, eight agreeing cases could all be agreement about an exclusion that never fires.

All 11 checks green on `ee3aa976d`, including the five required contexts.

pjt222:
## Round 1 applied — BLOCKING=1 SHOULD-FIX=3, all four fixed

The review ran its experiments in its own clone at the reviewed sha and every finding arrived with a repro. All four are applied at `778e6e216`.

### B1 — the wiring "proof" was a denylist of two identifiers. Correct, and it was the whole coverage claim for #872.

Measured by the reviewer: add `import { readdirSync } from 'node:fs';` as a **second** import line, revert one call site, plant a gitignored `scripts/local-probe.js` — the source scan stays green while the published count is wrong again. `check-readmes` cannot see it either, because it compares against a clean CI tree where a disk walk and a git listing agree by construction. In the other direction, changing the specifier to `'node:fs'` — the form every other file in this PR uses — reddened it *with the fix intact*.

Fix, following what #691 did for the functions that already moved: the three counts are now `scripts/lib/tree-counts.js`, taking `root`, and `scripts/test/tree-counts.test.js` asserts the **numbers** against a git fixture holding a gitignored file beside an untracked one. The source scan survives at the bottom of that file, renamed `TRIPWIRE (not the coverage claim)` and saying so in its own comment.

**The fix is proven against the reviewer's own scenario**, four arms, two of them controls, in a clone:

```
A  unmutated, clean                exit 0   pass 5
B  unmutated + the ignored file    exit 0   pass 5      <- does not flag any ignored file
C  mutated + the ignored file      exit 1   pass 3 fail 2   <- the B1 scenario
D  mutated, clean                  exit 1   pass 4 fail 1   <- fixture alone is enough
```

The first version of that envelope had the **wrong** control — it removed the planted file while leaving the mutation, which the fixture catches on its own, so it reported INCONCLUSIVE against a correct instrument. Worth stating because it is the same class of error the finding is about.

That envelope also caught a vacuous arm of my own in the same suite: the LIVE test asserted `security.includes(`${scripts} `)` — a digit run followed by a space, which matches somewhere in any version of that document. It now matches the published sentence, `**Scripts** (`scripts/`): N top-level`.

### S1 + S2 — the disk fallback is gone, and `insideWorkTree` with it

The reviewer measured what the fallback bought: a `git` failing the way `safe.directory` does (exit 128, "detected dubious ownership") put `tools/__pycache__` back into `notPlainFile` and a gitignored `evil.py` back into the inventory — silently, behind a `source: 'disk'` marker no consumer read. And its one named beneficiary does not exist: `scripts/` is in `REPO_ONLY`, `git grep` finds no shipped importer, and a package installed under a consumer's `node_modules/` sits **inside** their work tree where `node_modules` is ignored, so the fallback would have answered `[]` with certainty.

Every failure that is not git's documented "no path matched" now throws, with git's own stderr in the message — which also fixes **N1**, where `fatal: pathspec 'tools/empty' is beyond a symbolic link` was being reduced to "Command failed".

Consequence the reviewer measured and I had not: nine fixtures across `skills-inventory.test.js` and `tools-registry.test.js` were bare `mkdtemp` trees exercising that fallback rather than the production path. They are git repositories now, through a shared `scripts/test/_git-fixture.js` whose `cleanEnv` drops every `GIT_*` key and moves `HOME`/`XDG_CONFIG_HOME` — neither `cwd` nor `-C` is isolation.

### S3 — walk errors are no longer swallowed

Only `ENOENT`/`ENOTDIR` mean "nothing here"; `EACCES` and friends throw, with a test arm that skips under uid 0 rather than asserting vacuously. A tracked file under an unreadable directory used to vanish from the inventory, and git exits 0 with only `warning: could not open directory`, so nothing downstream would have refused on its behalf. `skills-inventory.js` makes exactly this argument three functions away about `executableFiles`.

### N4 — the perf number, restated for the current design

Removing `insideWorkTree` dropped 49 spawns per generator run. Same 46 call sites, same mount:

```
git-backed enumeration    3299 ms    (was 32612 ms with ls-files --others)
readdirSync baseline       173 ms
```

### Nine mutants, one per claim, all clean kills

| Mutation | File | Verdict |
|---|---|---|
| weaken the recursive ignore filter | `lib/git-files.js` | KILLED by 4 (`--allow-broad`, waiver named below) |
| disable the `topLevelEntries` skip | `lib/git-files.js` | KILLED by 2 |
| treat every git failure as "nothing ignored" | `lib/git-files.js` | KILLED by 1 |
| add `EACCES` to the absent-means-empty set | `lib/git-files.js` | KILLED by 1 |
| `topLevelEntries(root, TOOLS_DIR)` → `'scripts'` | `lib/tools-registry.js` | KILLED by 4 |
| `listNonIgnored(root, tree)` → `'i18n'` | `lib/skills-inventory.js` | KILLED by 8 |
| `scriptFileCount` → enumerate `workflows` | `lib/tree-counts.js` | KILLED by 2 |
| `workflowFileCount` → enumerate `scripts` | `lib/tree-counts.js` | KILLED by 2 |
| `localeTranslationCounts` → `localeRel` | `lib/tree-counts.js` | KILLED by 1 |

Every mutant parses and runs to completion, so no count is a crash signature. The one `--allow-broad` waiver is the first row: 4 of 10 is "broad for one line" by the tool's heuristic, and the four arms that die are the four whose assertion **is** the ignore rule.

`npm test` exit 0 (904 script tests, 141 CLI), `validate:integrity`, `check-readmes`, `check:generator-inputs`, `validate:line-endings` all exit 0, each run bare.

pjt222:
## Reweighted round 1: the S-rule finding, applied — my stated rationale was measured false

The reviewer's reweighted report arrived after the redesign and carried one finding the redesign had not answered. It is right, and it invalidates a sentence I had written in two places.

### "A gitignored file does not ship" is false

Measured in a clone at the reviewed sha — plant the #872 artefact and an untracked non-ignored sibling, then `npm pack --dry-run --json`:

```
packed files: 621
pycache in pack: [".../__pycache__/evil.py", ".../__pycache__/repro.pyc"]
fresh-untracked in pack: [".../fresh-untracked.py"]
```

With a `files` array and no `.npmignore`, **npm packs the working tree**. A local pack ships the ignored `.py`, the ignored `.pyc` and the untracked `.py` alike. The *release* pack contains none of them — because CI checks out a commit, and there the disk **is** the tracked set.

So "tracked or not ignored" describes neither artifact: it drops a `.pyc` a local pack ships and keeps an untracked `.py` the release does not.

### Two questions, two functions

| Consumer | Question | Function |
|---|---|---|
| `checkParity` (the tools gate) | what is in this working tree that git is not ignoring | `listNonIgnored` |
| `nonDocumentationFiles` (SECURITY.md) | what is in the committed artifact | `listTracked` |
| the translation counts | either rule serves; uses the gate rule | `topLevelEntries` |

`#830`'s acceptance criteria are why the gate keeps the first rule — an untracked new tool file with no registry row must stay visible. SECURITY.md takes the second, and the generated sentence now states its own basis rather than leaving the reader to infer it:

> All of it ships — **counted from the commit, which is what a release is packed from; a local `npm pack` packs the working tree instead and can include files this count excludes.**

Published count unchanged at 18. Nine fixtures follow the rule rather than the rule bending: files written after `initRepo` are untracked, so the inventory assertions go through a helper that commits first.

### W1 — a candidate is a pathspec, and `--literal-pathspecs` is not the escape

`git status --ignored` calls `tools/x*y.log` ignored; `check-ignore` reports it **not** ignored when a tracked sibling glob-matches the name. I measured the obvious fix and it does not exist:

```
$ git --literal-pathspecs check-ignore -z --stdin  <<< 'tools/__pycache__'
fatal: tools/__pycache__: pathspec magic not supported by this command: 'literal'
(exit 128)
```

Rejected for **ordinary** candidates too, so there is nothing to fall back to. A candidate carrying `*`, `?`, `[` or a leading `:` is now refused by name. No tracked path under the enumerated trees carries one today — a class, not a live case.

W2, W3 and W4 are recorded in the module header rather than guarded, because they are properties of asking git at all: `core.ignorecase` is `true` on this mount and `false` on a Linux runner; `core.excludesFile` and `.git/info/exclude` reach the answer; and a batch refuses as a unit, which is why the thrown message carries git's stderr.

### Three arms the review asked for by name

- **a symlinked directory** under the tree — a path *through* a symlink is exit 128 for the whole batch, and the `withFileTypes` walk is what stops one ever being generated. The arm pins that against a future refactor to a recursive or stat-following walk.
- **a `git` that exists and fails** — a PATH shim exiting 128 with "dubious ownership". The previous refusal arm used a bare directory, which answers a different question; this is the shape that silently restored both defects under the old fallback.
- **`listTracked` beside `listNonIgnored`** on one fixture where the two sets genuinely differ (a force-added ignored file, a tracked file, an untracked file).

### Mutants for the new code

| Mutation | Verdict |
|---|---|
| `listTracked(root, tree)` → `'i18n'` in the inventory | KILLED by 8 |
| `listTracked` asks `--others --exclude-standard` as well | KILLED by 2 |
| disable the pathspec-metacharacter refusal | KILLED by 1 |
| treat exit 128 as "nothing ignored" | KILLED by 2 |

`npm test` exit 0 (908 script tests, 141 CLI), `check:generator-inputs` exit 0.

One out-of-scope hazard from the same measurement is filed separately rather than fixed here: a local `npm publish` with a `__pycache__` present ships it, and `assertInterpretable` refuses the glob negation that would stop it.

PR #878, opened 2026-09-21
feat(skills): consult-a-decision-oracle — three gaps found by running the method in production

### LINKED ISSUE #875: skills: consult-a-decision-oracle — three gaps found by running the method in production (guard-before-pin, a reader for the logged model id, margin's preconditions)
## Summary

`skills/consult-a-decision-oracle` shipped in #871 and was then **run** — a peer production integration followed its steps against a live oracle and reported back three places where the method as written is incomplete. All three are method-level, apply to any scored classifier, and carry no vendor constant and no data from the reporting system. None is a correction of a claim the file makes; each is a gap beside one.

The change cannot be a one-line edit, which is why it is an issue rather than a commit: `skills/consult-a-decision-oracle/SKILL.md` is **491 lines against a gate-enforced English ceiling of 494** (500 minus the six provenance fields every mirror carries — `scripts/check-skill-line-ceiling.js`), so three lines of headroom. Adding the text below means an extraction pass into `references/EXAMPLES.md` first. And the skill has four byte-copy mirrors (de, zh-CN, ja, es) from #873, so the English edit strands them until `refresh:stubs` plus `--stamp` runs — which under the standing rule is its own reviewed change, not a rider.

## Findings

1. **[HIGH] Step 6 is a precondition of Step 5's pin, not a later refinement.** The file presents pinning the model id first ("pinning is right — but pin plus a tripwire") and guarding environment-sourced values afterwards. The ordering is backwards in the one case that matters: **pinning is the change that first puts a value into an environment-sourced variable.** If the config layer leaves an unset variable as literal placeholder text, every call after the pin fails with an unknown-model error, and the fail-open the file prescribes swallows it silently. Found because the reporting integration had exactly this shape — the model id read from the environment with no placeholder guard, twelve lines below a credential that had one — latent only because the variable was absent from its production config block.

2. **[HIGH] "Log the resolved model identifier" specifies no reader, and the tripwire covers only one direction.** Step 5 says to log the resolved id on every call, including agreements, and the specified alarm is for the pin-retired error — which can only fire in the **pinned** direction. The **floating** direction, which the file itself names as a silent failure, has no detector at all. Reported outcome: the id was logged on every call, for days, and nothing ever read it. Detectable in principle, undetected in practice. The symmetric detector is cheap: read the recorded id back out of the log and compare it against the id the operating point was measured on. It should not move any exit code — a floating model id is not the same question as "is the service up".

3. **[MEDIUM] "Try the top-two margin instead of the confidence scalar" has two unstated preconditions.** (a) Margin needs the full distribution, and a client that parses only the winning option has already discarded it — so it is a client change first and a measurement second, which is worth saying before someone reaches for it as a quick experiment. (b) Margin has **no analogue on a yes/no question type**, where the probability is the only axis; advice framed as "prefer margin" silently does not apply there. Related trap worth one line in the same place: a yes/no answer carries no confidence field, so a `confidence >= X` gate copied from a choice question **never fires** — `undefined >= X` is false, nothing logs an error, and it fails safe and silent.

## Acceptance criteria

- [ ] Step 5 states that every environment-sourced value in the request is guarded **before** the pin lands, and says why: the pin is what first puts a value in that variable
- [ ] The Step 5/6 checklist gains "every environment-sourced value in the request is guarded BEFORE pinning" beside the existing pin-retired alert item
- [ ] The floating direction has a specified detector: read the recorded model id back out of the log and compare it against the id the operating point was measured on; stated as not moving an exit code
- [ ] The margin advice names both preconditions — the client must parse the full distribution, and there is no margin on a yes/no question type
- [ ] The `undefined >= X` trap is stated where the confidence gate is defined, not only in the margin paragraph
- [ ] `skills/consult-a-decision-oracle/SKILL.md` is at **494 lines or fewer** after the change; the overflow goes to `references/EXAMPLES.md` under the progressive-disclosure pattern, not into an allowlist (#843)
- [ ] The four mirrors are refreshed and stamped in the same PR or an immediately following one: `npm run refresh:stubs -- skills consult-a-decision-oracle --verify`, then the refresh, then `--stamp <sha>`
- [ ] `npm run validate:skills` / `validate:integrity` green, and the line-ceiling check run over **English and every mirror**, not English alone
- [ ] No duration is attached to finding 2's "nothing ever read it" — the first figure reported was wrong by an order of magnitude and was corrected in review, not by the reporting process, which is the finding one layer up: a number in prose deserves the same provenance check as a number in code and almost never gets it

## Context

Reported 2026-09-21 by the `agent-almanac-companion` peer session, which holds the production integration. Its repository is **private**: no issue number, corpus row or measured figure of its own is reproduced here or belongs in the skill. Every item above is method-level and was cleared by that session as safe to publish as written.

Two things deliberately **not** carried into the skill:

- That integration decided **not** to pin, having built the detector first and left the choice open. That is a local call under the file's own argument that both directions fail silently in opposite ways; the file's "pinning is right" survives it and is not being softened.
- The vendor-example threshold the file already refuses to quote stays unquoted, for the reason it gives.

Related: #871 (the skill), #873 (the four mirrors).

pjt222:
Closes #875.

## Summary

A peer session followed `skills/consult-a-decision-oracle` against a live oracle and reported three places where the method as written is incomplete. All three are method-level, apply to any scored classifier, and carry no vendor constant and no data from the reporting system. None corrects a claim the file makes — each sits beside one.

### 1. Step 6 is a precondition of Step 5's pin, not a later refinement

The file presented pinning first and guarding environment-sourced values afterwards. That ordering is backwards in the one case that matters: **pinning is the change that first puts a value into that variable.** A configuration layer that leaves an unset variable as literal placeholder text then fails every call with an unknown-model error, and the fail-open this skill prescribes swallows it.

Found because the reporting integration had exactly that shape — the model id read from the environment with no placeholder guard, twelve lines below a credential that had one — latent only because the variable was absent from its production config block.

### 2. "Log the resolved model identifier" specified no reader

Step 5 says to log the resolved id on every call, and the alarm it prescribes is for the pin-retired error — which can only fire in the **pinned** direction. The **floating** direction, which the same step names as a silent failure, had no detector at all. Reported outcome: the id was logged on every call, for days, and nothing ever read it.

The symmetric detector is cheap and is now specified: read the recorded id back out of the log and compare it against the id the operating point was measured on. It should not move an exit code — a floating model id is not the same question as "is the service up".

**No duration is attached to that claim.** The first figure reported was wrong by an order of magnitude and was corrected in review, not by the reporting process — which is the finding one layer up: a number in prose deserves the same provenance check a number in code gets, and almost never gets it.

### 3. The margin has two unstated preconditions

In `references/EXAMPLES.md`, beside the existing scalar-choice material:

- it **needs the full distribution**, and a client that parses only the winning option has already discarded it — so "try the margin instead" is a client change first and a measurement second;
- it has **no analogue on a yes/no question type**, where the probability is the only axis, and the question type is usually chosen per call site rather than per service.

Plus the related trap that fails **safe and silent**: a yes/no answer may carry no confidence field, so a `confidence >= X` gate copied from a choice question never fires — the comparison is false, nothing is raised, and the gate reports as one that simply never had cause to act.

**That trap is also stated where the gate is defined**, in Step 4, which is what #875's AC 5 asks for and what the first revision of this PR missed: a reader who copies the Step 4 gate onto a yes/no call site never reaches the margin paragraph. Measured portably rather than asserted for one language — node gives `undefined >= 0.5 === false`, python3 raises `TypeError` on `None >= 0.5`, and a language that raises is failing the better way.

## The line budget, which is why this is not an append

`SKILL.md` was at 491 against a gate-enforced English ceiling of **494** (500 minus the six provenance fields every mirror carries). The additions are paid for rather than appended — seven extractions after two review rounds:

- the Step 6 boolean paragraph, **duplicated verbatim** in `EXAMPLES.md`, is now a pointer to it;
- two Common Pitfalls keep their claims while their worked numbers move out (`#what-a-confidence-score-cannot-tell-you`, `#how-small-a-surgical-gate-is`);
- the authoring-sessions story, the 39-day corpus accident, the live-by-default wiring block and Step 3's full grading table follow them;
- several blocks tightened without dropping a claim — and where a reflow DID drop one, the review caught it twice and both are restored.

**English lands at 491 and the mirrors at 497.** The first revision of this PR sat at exactly 494 and 500, which the review measured as *zero* headroom: one appended line would have reddened English and all four mirrors at once. Four lines of margin is deliberate, not spare capacity.

## Mirrors

The four mirrors (de, zh-CN, ja, es) are byte-copy scaffolds, so an English edit strands them. Refreshed and stamped in their own two commits, per the documented recipe:

```
npm run refresh:stubs -- skills consult-a-decision-oracle --verify   # 4 diverged at line 51
npm run refresh:stubs -- skills consult-a-decision-oracle            # 4 written
git commit -- i18n/
npm run refresh:stubs -- skills consult-a-decision-oracle --stamp <that commit>
```

`translation_date` is unchanged: for a stub that field records when it was first scaffolded, and no translation event happened here.

## Verification

```
npm test                       EXIT=0   (912 script tests, 141 CLI)
npm run validate:integrity     EXIT=0
node scripts/check-i18n-fence-parity.js --id consult-a-decision-oracle   EXIT=0
```

Each run bare and read from `$?`. `validate:translations` reports the corpus-wide backlog (2577 stale across 11 locales); `consult-a-decision-oracle` does not appear in it.

The line-ceiling check was run over **English and every mirror**, not English alone — checking the English file by itself is what let a 500-line file pass locally and fail CI on four mirrors in #843.

## Not carried, deliberately

The reporting integration decided **not** to pin, having built the detector first and left the choice open. That is a local call under this file's own argument that both directions fail silently in opposite ways; "pinning is right" survives it and is not softened here. The vendor threshold this skill already refuses to quote stays unquoted.


pjt222:
## Round 1 applied — BLOCKING=1, SHOULD-FIX=6, NOTE=7, all of them

### B1 was an acceptance criterion I wrote and then did not meet

#875 AC 5: the `undefined >= X` trap must be stated **where the confidence gate is defined**, not only in the margin paragraph. It was only in `EXAMPLES.md`, and a reader who copies the Step 4 gate onto a yes/no call site never gets there. It now sits two lines after the gate operator is defined, in the same class as the sentence above it — a strict `>` on a corpus scoring 1.00 fires on nothing; a missing scalar disables the gate the same silent way.

### SF2 is the one worth reading

The guard-before-pin paragraph justified itself with "fails every call with an unknown-model error, and fail-open swallows it" — two sentences after the step mandates a tripwire on the pin-retired error. The review put it as a fork, and both branches cost the paragraph its justification:

- same error class → the mandated tripwire fires, so the paragraph describes a pin **without** its tripwire, which the step forbids;
- different class → the file is asserting a vendor error taxonomy its own preamble promises not to pin.

The justification that survives both is Step 6's own: **the guard refuses the value beforehand; a tripwire tells you afterwards, and only if that is the error it watches.**

### The rest

| | |
|---|---|
| SF1 | "for days" deleted — a duration this file's own Pitfall 5 forbids in prose, from a system no reader can inspect, whose first reported figure was already off by an order of magnitude |
| SF3 | Expected listed five things, On-failure branched on four |
| SF4 | the reader was specified only by what it must *not* do; Step 2's "recording without a reader is not monitoring" applies one level up, so a mismatch must reach a person |
| SF5 | the "suite results, not graded production rows" qualifier my extraction dropped |
| SF6 | three words compression weakened or lost, restored |
| N2–N7 | log instruction moved ahead of the paragraphs citing it; stale "Three parts"; a `[reported]` provenance class, since relayed-by-a-peer is neither `[shipped]` nor `[inference]`; dangling clauses |

**N6, measured rather than restated:** node gives `undefined >= 0.5 === false`, python3 raises `TypeError` on `None >= 0.5`. For a language-agnostic skill the portable statement is that a missing field comparing false and raising nothing is the dangerous case — a language that raises is failing the better way.

**One thing the review's suggestion would have broken.** The provenance tags went into section **bodies**, not headings: `— [shipped]` in a heading changes its GitHub slug and would have broken two cross-references the same review had just verified. All eight anchors resolve.

### Paying for it was the work

The review measured the margin as exactly zero — one appended line reddens English at 495 **and** all four mirrors at 501 — and asked for ≤492 so the next edit need not repeat the exercise.

```
English 490 -> +1 = 491   (ceiling 494)
  de/es/ja/zh-CN 496 -> +1 = 497   (mirror limit 500)
```

Four lines of headroom, from zero. The authoring-sessions story, the 39-day corpus accident, the live-by-default wiring block and the three-denominator numbers now live in `EXAMPLES.md` behind pointers.

**N1 is deliberately not applied.** Swapping Steps 5 and 6 removes the forward dependency, but renumbers cross-references in `EXAMPLES.md`, the Related Skills section and any external pointer — and the precondition is now stated in both directions, which was the readability problem. Say the word if you disagree.

```
npm test                                                  EXIT=0   (912 script tests, 141 CLI)
npm run validate:integrity                                EXIT=0
node scripts/check-i18n-fence-parity.js --id consult-a-decision-oracle   EXIT=0
```

Three commits, in the cascade shape the review specified: one English batch, one refresh, one stamp.

PR #879, opened 2026-09-21
feat(scripts): refuse a pack whose contents the published commit does not carry

### LINKED ISSUE #876: packaging: a local `npm publish` ships a gitignored `__pycache__`, and the negation that would stop it is refused
## Summary

With a `files` array in `package.json` and no `.npmignore`, **npm packs the working tree** — it does not honour `.gitignore`. So a maintainer who has imported a Python asset (which writes a `__pycache__/` beside it) and then runs `npm publish` from that tree ships the bytecode, and would ship any other gitignored file under a shipped tree.

Measured during the #874 review, in a clone at `b8eee5b7e`, with the #872 artefact and an untracked sibling planted under `skills/consult-a-decision-oracle/references/`:

```
$ git check-ignore -v .../__pycache__/evil.py
.gitignore:70:__pycache__/	skills/consult-a-decision-oracle/references/__pycache__/evil.py
$ git status --porcelain | rg fresh-untracked
?? skills/consult-a-decision-oracle/references/fresh-untracked.py
$ npm pack --dry-run --json
packed files: 621
pycache in pack: [".../__pycache__/evil.py", ".../__pycache__/repro.pyc"]
fresh-untracked in pack: [".../fresh-untracked.py"]
```

The **release** is not affected: `release.yml` packs from an `actions/checkout` of a commit, where the working tree is the tracked set. This is the local-publish path only.

## Why the obvious fix is not available

The natural repair is a negation in `files` — `"!**/__pycache__/"`. `scripts/lib/skills-inventory.js`'s `assertInterpretable` **refuses** it, and correctly: it models npm's negations as "trailing slash means prefix, otherwise exact path", and a glob would be expanded by npm and compared literally by that module, so the two would disagree in silence. Adding the negation without teaching the module the pattern would make the published file count wrong in the quiet direction.

## Suggested shape

A `prepack` guard that refuses when the tree carries anything ignored under a shipped directory:

```bash
git status --porcelain --ignored -- $(node -p "require('./package.json').files.filter(f=>!f.startsWith('!')).join(' ')")
```

Non-empty output means the pack would differ from the release. Exiting non-zero there turns a silent difference into a refusal at the moment it matters.

Two things to get right, both learned in #874:

- **`prepack` must not be added blindly.** `scripts/lib/skills-inventory.js` asserts this package declares no install-time hooks, and `INSTALL_HOOKS` is `preinstall`/`install`/`postinstall`/`prepare`. `prepack` is not one of those — it runs in the PUBLISHER's tree, not a consumer's — but the inventory sentence should be checked against whatever is added.
- **Prove it can fail**: create the artefact, confirm the guard refuses, remove it, confirm the guard passes. A guard that cannot fail is the failure mode this repository keeps finding in its own gates.

## Acceptance criteria

- [ ] A `prepack` (or equivalent) guard refuses a publish from a tree carrying ignored files under a shipped directory
- [ ] The guard is proven able to fail: red with `__pycache__` present, green without, both recorded
- [ ] `npm pack --dry-run --json` on a contaminated tree shows the artefact excluded, or the publish refused
- [ ] The inventory claim in `SECURITY.md` about install-time hooks is re-checked against whatever hook is added
- [ ] If a `files` negation is chosen instead, `assertInterpretable` is taught the pattern in the same commit — measured against `npm pack --dry-run`, never reasoned about

## Context

Found by the adversarial review on #874 while checking the premise under that PR's rule. The PR itself is unaffected once merged: `SECURITY.md`'s inventory now counts the **commit**, and says so, precisely because the working tree and the release can differ. This issue is the other half — the publish path where that difference becomes real bytes.

pjt222:
Closes #876.

## The defect

**npm packs the working tree** under a `files` array — it does not honour `.gitignore`. Measured during the #874 review, in a clone carrying the #872 artefact plus an untracked sibling:

```
packed files: 621
pycache in pack: [".../__pycache__/evil.py", ".../__pycache__/repro.pyc"]
fresh-untracked in pack: [".../fresh-untracked.py"]
```

A maintainer who has imported a Python asset — which writes a `__pycache__/` beside it, and is how **both** #868 and #872 were produced — and then runs `npm publish` from that tree ships the bytecode. The **release is unaffected**: `release.yml` packs from an `actions/checkout` of a commit, where the working tree *is* the tracked set. This is the local-publish path only, and it is silent: nothing in `npm publish` prints what it decided not to exclude.

## The guard

`scripts/check-publishable-tree.js`, wired at **`prepack`**, refuses three classes named separately because they have different fixes:

| Class | Fix |
|---|---|
| `IGNORED` — a path git ignores under a shipped directory | `rm` it, or publish from a clean clone |
| `UNTRACKED` — a path git does not know about | commit it or remove it |
| `MODIFIED` — a tracked path whose working bytes differ from the commit | commit or revert it |

Refusing one without the others would leave most of the hole open — npm packs the disk in every case. The third class came from the review, which measured 14 bytes packed against 7 committed while this check was reporting "a pack here matches the commit".

**Scoped to `package.json`'s own `files` — both halves.** Divergence under `scripts/` or anywhere else that does not ship is not the subject; a guard that refused every publish from a working checkout would be disabled within a week. The **negations** matter too: content under `skills/_template/` can never reach a tarball, so refusing it would block a publish over a file that cannot ship. A pathspec cannot express that — `:(exclude)` does not drop the ignored-DIRECTORY entry git reports for a wholly-ignored tree — so the results are filtered, with the same exact-versus-prefix rule `skills-inventory.js` uses. Tests assert both directions and both mutants die.

## Why not the obvious `files` negation

A negation matching `__pycache__` at any depth is the natural repair, and `scripts/lib/skills-inventory.js`'s `assertInterpretable` **refuses** it — correctly. That module models npm's negations as "trailing slash means prefix, otherwise exact path"; npm expands globs; the two would disagree in silence about the published file count, which is the quiet direction. Teaching the module that pattern needs its own `npm pack --dry-run` measurement, which the issue said and this does not shortcut.

## A correction the new tool produced on its first real use

`tools/mutation-envelope.sh` (merged in #874) reported one row not a clean kill: swapping `--ignored=matching` for plain `--ignored` changed **no test**. The docstring claimed that flag was what listed a nested ignored file individually. Measured on git 2.43 over the three shapes this check meets, that is false with `-uall` set:

```
with -uall      matching and traditional agree on the VERDICT; they differ only in whether a
                wholly-ignored directory prints as `__pycache__/` or as the `.pyc` inside it
without -uall   an untracked file collapses to its DIRECTORY. For the `__pycache__` case,
                traditional prints the non-ignored PARENT only when that parent holds no
                tracked file; with a tracked sibling every form prints the `__pycache__/`
                entry itself
```

So **`-uall` is the load-bearing flag** and `--ignored=matching` is a precision choice. The fixture now pins the reported **path**, not only the refusal: an operator told a directory when one file is the problem searches in the wrong place. The comment says what was measured rather than what was assumed.

## SECURITY.md

The generated inventory now names the hook, and the sentence it rests on stays true: `prepack` runs in the **publisher's** tree when packing, never in a consumer's on install. A test asserts that membership against `INSTALL_HOOKS` rather than trusting the prose — if `prepack` ever joined that list, the sentence would become false and `assertInventoryClaims` would throw.

## Round 2, and what it cost

A second adversarial round returned BLOCKING=1, SHOULD-FIX=6, NOTE=9. The blocking item — a staged rename whose old side sits under a directory negation — turned out to be **already closed by accident**: it was measured at a sha predating the `-z` change, and `-z` splits the rename into a record carrying the *new* path, which is what gets tested. `--no-renames` is applied anyway, because it removes the question instead of depending on a record shape and makes the answer independent of `status.renames` config.

Three findings from that round were defects the first round's fixes introduced or missed:

- **`carvedOut` was order- and re-include-blind**, and both shapes fail open. Re-derived rather than taken on report, because it contradicts a measurement already in `skills-inventory.js`: a **directory** negation placed before its inclusion is *dead* (npm packs the directory), while a **file** negation in that position is honoured — so the existing "ordering was refuted" note is right for the shape it measured and wrong in general. `assertInterpretable` refuses both bad shapes now, and the guard delegates to `shippedEntries` so a manifest cannot pass at prepack while failing the inventory.
- **My subprocess test passed at both arms via a crash.** The script was copied to `<fixture>/check.mjs`, making ROOT the fixture's *parent*, so `readFileSync` threw ENOENT and the process exited 1 having checked nothing — the plain arm passed with the *old* guard too. It asserts `REFUSED` on stderr now, with a clean-tree arm asserting exit 0, which a crash can never produce.
- **The pack-hook row was measured under a snapshot gate.** `check-readmes` cannot tell a derivation from a literal rendering the same bytes: reverting to a hardcoded `['prepack']` regenerated SECURITY.md byte-identically and the mutant survived. `packHookSentence(pkg)` is a pure function now, unit-tested with a manifest lacking the hook.

And one of my own, caught by reproducing the reviewer's lab: the B1 fix it supplied was `process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href`, and I typed it without the first half. `pathToFileURL(undefined)` **throws**, so a fix for a fail-open introduced a fail-closed two lines away — the module crashed on import from `node -e`, a REPL or a loader. Every other test imports it normally, so nothing could see it.

Four guards added during that round then survived their own mutants, which is the same class one level down; each has a test now.

## Verification

Proven able to fail by hand on this tree before the tests existed, and the clean tree still passes:

```
$ mkdir -p skills/consult-a-decision-oracle/references/__pycache__ && printf 'x' > .../repro.pyc
$ node scripts/check-publishable-tree.js
REFUSED: 1 IGNORED path(s) under a shipped directory would be packed and are not in the commit — remove them, or pack from a clean clone:
  !! skills/consult-a-decision-oracle/references/__pycache__/

$ printf 'scratch\n' > skills/consult-a-decision-oracle/scratch.txt
REFUSED: 1 UNTRACKED path(s) under a shipped directory would be packed and are not in the commit — commit or remove them:
  ?? skills/consult-a-decision-oracle/scratch.txt

$ node scripts/check-publishable-tree.js   # clean tree
EXIT=0
```

**Seventeen mutants, each under the command that owns its claim** — via `tools/mutation-envelope.sh` and the plan file it takes.

| Mutation | Command | Verdict |
|---|---|---|
| main-module guard → `file://${process.argv[1]}` template | `test:scripts` | KILLED by 1 |
| drop the `process.argv[1] &&` half of that guard | `test:scripts` | KILLED by 1 |
| stop classifying `!!` / `??` / the rest | `test:scripts` | KILLED by 3 / 4 / 1 |
| disable the `files`-negation carve-out | `test:scripts` | KILLED by 1 |
| make every negation a prefix match | `test:scripts` | KILLED by 1 |
| accept a directory negation before its inclusion | `test:scripts` | KILLED by 1 |
| accept a re-inclusion under a negated prefix | `test:scripts` | KILLED by 1 |
| drop `--no-renames` | `test:scripts` | KILLED by 2 |
| drop `-uall` | `test:scripts` | KILLED by 1 |
| drop the `-- <shipped paths>` pathspec | `test:scripts` | KILLED by 1 |
| let an absent `files` array pass | `test:scripts` | KILLED by 1 |
| stop skipping npm's default excludes (names) | `test:scripts` | KILLED by 1 |
| stop skipping them (patterns) | `test:scripts` | KILLED by 1 |
| add `node_modules` to the default-excludes set | `test:scripts` | KILLED by 1 |
| return empty instead of throwing on a git failure | `test:scripts` | KILLED by 1 |
| revert the pack-hook sentence to a literal | `test:scripts` | KILLED by 1 |

**One row is deliberately absent, and saying why is the point.** Dropping `-z` kills **nine** tests, and that count is not coverage: the parser splits on NUL, so removing the flag turns the whole output into a single record and everything downstream fails at once — the broad-kill shape `CLAUDE.md` § Proving a Gate Can Fail documents (15 versus 1). The flag and the parser have to change together, so no single-line mutant can measure it honestly. The discriminating arm is the quoting test, which is the one test that fails for the reason `-z` exists.

**Four mutants survived along the way, and every one was a real gap.** Prefix-matching a file negation (a `.bak` sibling carved out silently). The pack-hook row, surviving a suite that does not run the gate owning it — now a pure function unit-tested instead. And, in the commit that was *fixing* that class, four new guards with no test that could see any of them: the two `assertInterpretable` refusals, `--no-renames`, and the argv guard.

**`--no-renames` needed the right fixture twice.** A rename *within* a shipped directory kills the mutant but exercises no carve-out; the direction that produces a wrong **answer** is a rename *into* the negated directory, where `-z` alone carves out the new path, the parser drops the origin record, and the guard reports OK while the pack lacks a committed file. Both directions are fixtures now.

```
npm test                     EXIT=0   (934 script tests, 141 CLI)
npm run validate:integrity   EXIT=0
npm run check-readmes        EXIT=0
```


pjt222:
## Round 1 applied — BLOCKING=1, SHOULD-FIX=5

### B1: the guard was fail-open, which is the one direction a guard must not fail

The main-module check compared `import.meta.url` against a `` `file://${process.argv[1]}` `` template. `import.meta.url` percent-encodes; `process.argv[1]` does not. At a path containing a space the two never match, the body never runs, and `prepack` exits 0 having checked nothing.

Reproduced before fixing, both arms, same fixture with a gitignored `.pyc` staged to ship:

```
/tmp/…/plain        node check.js exit=1   (refuses, correct)
/tmp/…/with space   node check.js exit=0   (fail-open)
```

`pathToFileURL(process.argv[1]).href` closes it; both arms exit 1. A test drives the script as a subprocess from **both** path shapes, so the regression has a name.

### The five should-fixes

| | Finding | What changed |
|---|---|---|
| S1 | a modified/staged/deleted tracked file packs its **working-tree** bytes — measured 14 packed against 7 committed — while the OK line said "a pack here matches the commit" | `MODIFIED` is a third class with its own remedy; the OK line now says "every shipped path is tracked and unmodified" |
| S2 | a `files` negation was dropped from the scan but not the results, so content under `skills/_template/` — which npm never packs — was refused | results are filtered through the negations. A pathspec cannot do this: `:(exclude)` does not drop the ignored-**directory** entry git reports for a wholly-ignored tree |
| S3 | the docstring's "traditional prints the non-ignored parent" does not reproduce for a **tracked** parent | measured both ways and stated with its condition — it holds only when the parent holds no tracked file |
| S4 | `It does declare prepack` was an unconditional literal; deleting the hook left `check-readmes` green | derived from `package.json`, and proven by its own negative control: removing the hook now makes the check exit 1 |
| S5 | a `--check` flag the docstring named and the code never implemented | removed; the script takes no arguments |

### Two mutants survived, and both were real

The envelope over the new surface returned 8 of 10 clean:

- **`negation-overbroad`** — making every negation a prefix match changed no test. Prefix-matching a *file* negation carves out every sibling whose name merely starts with it (`SKILL.md.bak` beside an exactly-negated `SKILL.md`), and those pack. Now pinned; the mutant dies.
- **`pack-hook-derived`** — surviving `npm run test:scripts` because that suite does not run `check-readmes`, which is the gate that owns the claim. Re-run against `npm run check-readmes`: `MUTANT KILLED`. The row is recorded under the command that can kill it rather than the one that cannot.

Final table, each row under the command that owns it:

| Mutation | Command | Verdict |
|---|---|---|
| main-module guard → template literal | `test:scripts` | KILLED |
| stop classifying `!!` / `??` / other | `test:scripts` | KILLED by 3 / 4 / 1 |
| disable the negation carve-out | `test:scripts` | KILLED by 1 |
| make every negation a prefix | `test:scripts` | KILLED by 1 |
| drop `-uall` | `test:scripts` | KILLED by 1 |
| drop the `-- <shipped>` pathspec | `test:scripts` | KILLED by 1 |
| return empty instead of throwing on git failure | `test:scripts` | KILLED by 1 |
| stop deriving the pack-hook sentence | `check-readmes` | KILLED |

```
npm test   EXIT=0   (923 script tests, 141 CLI)
```

pjt222:
## The full report arrived — eight NOTEs applied, and two of them were defects

The harness lost the long report; it came back in four parts. Everything in it is now applied. Two items were not notes at all.

### N7 — a SECOND fail-open, in the same file as B1

With no `files` array the check returned `{ ignored: [], untracked: [] }` and printed *"a pack here matches the commit"*, while npm packed three untracked files. That is exactly the configuration where npm honours the root `.gitignore` for you and **only this check** could catch the UNTRACKED class. An empty inclusion list throws now.

Two fail-opens in one guard, found by one review. Both had the same shape: a branch that returns "nothing to see" for a condition nobody had asked what it meant.

### N2 — the report was naming paths npm never packs, and the fix nearly blinded it

Measured: `.npmrc`, `.DS_Store` and a `.orig` file are absent from `npm pack --dry-run --json` regardless of `files`. Refusing them is a false statement that blocks a publish over a Finder dropping. Skipped now.

**`node_modules` is deliberately not in that set**, and that is the part worth reading: the same report measured `skills/real/node_modules/dep/index.js` **present** in the listing. A nested `node_modules` under a shipped directory *does* ship, so the guard's refusal of it is real rather than redundant with npm's root exclusion. Adding it to the exclusion set — which is what "npm ignores node_modules" would have led me to do — would have blinded the check to a real case while fixing a false one.

### N5 — the quoting reached the report verbatim

Porcelain quotes a path carrying a space or a non-ASCII byte: `"skills/real/umlaut-\303\244.md"`. `-z` now, and the parse splits on NUL. The record filter also drops a rename's ORIGIN record, which is not a path npm would pack.

### S2's mechanism, properly this time

`isExcludedFromPackage` is **exported** from `skills-inventory.js` and imported here, rather than re-typed as it was in the first fix. Two copies of an accept rule is the drift `CLAUDE.md` § Excluding a Template is about, and the review asked for the export by name.

### N3, N4, N6 — three measured consequences of the hook, in the docstring

- `npm pack --ignore-scripts` bypasses it entirely — worth knowing, since that is the common reflex when a hook fails.
- Publish hook order is `prepublishOnly` → `prepack` → `prepare` → `postpack`, so a future build-on-`prepare` artefact would land **after** this check.
- Declaring `prepack` at all changes consumer-side behaviour on a git install, though the script never runs there: pacote's `lib/git.js` treats `prepack` as a key that makes a clone worth preparing, so `npm install pjt222/agent-almanac` now runs a nested `npm install --force` inside the temp clone. Measured across twelve install shapes; the hook itself fired on none of the consumer installs.

### Five more mutants, all clean

| Mutation | Verdict |
|---|---|
| let an absent `files` array pass | KILLED by 1 |
| drop `-z` | KILLED by 8 |
| stop skipping npm's default excludes | KILLED by 1 |
| add `node_modules` to the default-excludes set | KILLED by 1 |
| make the negation filter a no-op | KILLED by 1 |

```
npm test   EXIT=0   (927 script tests, 141 CLI)
```

pjt222:
## Correction: two findings from the round-3 review shipped in this merge

The adversarial round-3 report on this PR arrived **after** the merge and graded `91c4b9515`,
one commit behind the merged head `950e35f2b` — which added only the rename-into test and nine
default-exclude names. So two of its SHOULD-FIX findings and two NOTEs were not addressed here
and landed on `main` as `c52d074f1`.

Re-measured against the merge commit before acting, all four reproduce:

- **`AD`, `MD` and `DD` are classified wrongly.** `report()` runs `.trim()` on git's two-column
  porcelain code and compares the pair whole, so it only ever matches a single-column code. With
  ` T`, `AD` and `MD` under a shipped directory, `npm pack --dry-run` lists `package.json` alone
  while this guard files `AD` and `MD` under MODIFIED — "would be packed with their WORKING-TREE
  bytes" — for files the pack carries no bytes of. `DD` (measured from a rename/rename conflict
  on git 2.43) matched none of the unmerged arms and fell through to MODIFIED too, telling an
  operator to "commit or revert" a path that is mid-conflict.
- **The `T` half of ABSENT-OR-RETYPED has no fixture**, which is why `absent-drops-T` survived
  its mutant: the only deletion test `git rm`s, and that emits `D `.
- **`assertInterpretable` refuses an interpretable array.** `["!lib-extra/x/", "lib"]` is refused
  as a directory negation placed before an inclusion it carves from, because the prefix test is
  on bytes rather than on a segment boundary; npm packs neither path from the other.
- **The `assertInterpretable` test leaks its `mkdtempSync` directory** — no `t.after`.

Fixed in **#883**, with a mutant per claim under `npm run test:scripts`.

Recording it here rather than only in the follow-up, because this PR's body describes the guard's
classification as settled and it was not.

Issue #621, opened 2026-08-16
mutation-check counts a ReferenceError crash as a KILL — a blind spot one level above its INVALID gate


pjt222:
`scripts/mutation-check.js` syntax-checks each mutant and reports `INVALID` when it does not parse, because CLAUDE.md records that "deleting a line carrying a brace makes the file fail to load, and node:test reports that as one failing test — indistinguishable from a real kill."

There is a second, quieter version of the same trap that the syntax gate cannot see: **a mutant that parses but crashes at runtime.**

## Observed

While covering #552 (PR #615), deleting a variable declaration:

```
npm run mutation-check -- --file scripts/normalize-i18n-fences.js \
  --delete-matching 'const stamped = stampFrontmatterField(repairedText, FENCE_BASIS_FIELD, basisLabel);' \
  --test 'npm run test:scripts'
```

reported **`MUTANT KILLED by 15 failing test(s)`**. But the surviving `if (stamped !== null) { … }` then references an undeclared binding, so under ESM strict mode every test that *executes* that module throws `ReferenceError`. The 15 failures prove the code is *reached*, not that its *effect* is asserted. The mutant parses cleanly, so the `INVALID` gate passes it through and the tool reports a strong-looking kill.

The honest instrument for the same line was a semantic mutation — reverting a condition to a weaker one — which parses, runs to completion, and died to exactly 1 test: the one written for it. Same line, same tool, 15 versus 1, and only the 1 means anything.

This matters because the kill count is quoted in commit messages and PR descriptions as evidence of coverage strength. A 15-kill reads as far stronger than a 1-kill, when here it was the reverse.

## Why it is hard to fix perfectly, and a cheap partial

A general "did the mutant crash rather than fail an assertion" check is not decidable from the exit code alone. But two cheap signals would catch most instances:

1. **Flag suspiciously broad kills.** If a mutant fails a large fraction of the suite — or fails tests in files unrelated to the mutated module — report `SUSPECT: broad failure, likely a crash rather than a behavioural kill` alongside the count, rather than a clean `KILLED`.
2. **Grep the test output for `ReferenceError` / `TypeError: … is not a function` / `Cannot access … before initialization`** and downgrade to `SUSPECT` when the failures are dominated by those. `mutation-check.js` already parses node:test output for the `fail N` count, so it has the text in hand.

Either keeps the tool's central promise — that a reported KILL means "a test asserts this behaviour" — which a crash does not establish.

## Acceptance criteria

[...]

PR #883, opened 2026-09-21
fix(scripts): read git's WORKTREE column, and prefix a negation on segments (#879 round 3)

pjt222:
The #879 review's round-3 report arrived after the merge and graded `91c4b9515` — one commit
behind the merged `950e35f2b`, which only added the rename-into test and nine default-exclude
names. Its two SHOULD-FIX findings and two of its NOTEs were therefore still live on `main`.
Each was re-measured against the merge commit `c52d074f1` before being touched.

## SF1 — the predicate read a trimmed pair, and npm packs the WORKTREE

git's porcelain code is two columns, index then worktree. `byRemedy` ran `.trim()` on the pair
and compared it whole, so it matched single-column codes only. Measured on `c52d074f1` with
` T`, `AD` and `MD` under a shipped directory:

```
=== git says ===
 T skills/real/SKILL.md
AD skills/real/ad.md
MD skills/real/keep.md

=== what npm packs ===
package.json

=== what the merged guard reports ===
REFUSED: 1 ABSENT-OR-RETYPED path(s) … the pack LACKS a file the commit has
   T skills/real/SKILL.md
REFUSED: 2 MODIFIED path(s) … with their WORKING-TREE bytes, not the committed ones
  AD skills/real/ad.md
  MD skills/real/keep.md
```

npm packed none of the three. Describing `AD` and `MD` as packed with their working-tree bytes is
a false statement about files the pack carries no bytes of, and it sends the operator to `git
checkout` when the fix is `git restore` or a commit.

The predicate now reads the **worktree column**: `D`/`T` there under any index column, plus
`D `/`T ` where the worktree agrees with an index that already dropped or retyped the path.

## SF1b — `DD` fell through to MODIFIED

The unmerged arm matched `UU`, `AA`, `startsWith('U')` and `endsWith('U')`, which is six of the
seven unmerged codes. `DD` matched none of them. The reviewer flagged it as unmeasured, so it was
measured rather than reasoned about — a rename/rename conflict on git 2.43:

```
DD skills/real/SKILL.md
UA skills/real/left.md
AU skills/real/right.md
```

A reachable shape, and the old code told the operator to "commit or revert" a path that is
mid-conflict. `DD` joins the set, and the unmerged arm is tested **first** so the worktree column
cannot claim `DD`, `DU` or `UD` from it.

## SF2 — the `T` half had no fixture

`absent-drops-T` survived the round-3 envelope: the only deletion fixture `git rm`s, which emits
`D `, so nothing ever exercised the typechange half of the class it shares a predicate with. The
new fixture builds all three shapes — `git add` then `rm` (`AD`), an edit staged then `rm` (`MD`),
and `rm f && ln -s /dev/null f` (` T`) — and asserts the codes, not just the count, so it cannot
pass while reporting something else.

## Q1 — a directory negation prefixed on bytes, not on segments

`assertInterpretable` refused `["!lib-extra/x/", "lib"]` as a directory negation placed before an
inclusion it carves from. `'lib-extra/x/'.startsWith('lib')` is true; npm packs neither path from
the other, so the array is interpretable and the refusal was false. The prefix is now taken on a
segment boundary, and the real refusal is still asserted with and without the inclusion's
trailing slash.

## Q3 — the `assertInterpretable` test leaked its fixture

It took `async ()` rather than `async (t)`, so it had no `t.after` and left a `publishable-files-*`
directory in `/tmp` on every run. It now tears down with `rmTree`, per CLAUDE.md's rule.

## Mutants

Four claims, four mutants, all under the command CI runs (`npm run test:scripts`), via
`tools/mutation-envelope.sh`:

| mutant | verdict |
|---|---|
| `worktree-column-reverts-to-trim` | MUTANT KILLED by 1 failing test |
| `absent-drops-T` | MUTANT KILLED by 1 failing test |
| `unmerged-drops-DD` | MUTANT KILLED by 1 failing test |
| `negation-prefix-loses-segment-boundary` | MUTANT KILLED by 1 failing test |

One failing test each — the honest shape for this instrument, not a broad kill that would mean
the mutant crashed on import (#621).

## Gates

Run bare from the repository root, reading `$?` rather than piping:

```
npm run test:scripts   → 937 pass, 0 fail, exit 0
npm run validate:integrity → PASSED, exit 0
npm test               → exit 0 (integrity, check-readmes, test:scripts, test:cli)
node scripts/check-publishable-tree.js → OK, exit 0
```

## What is NOT changed

The reviewer's remaining NOTEs concern the #879 PR body against the sha it graded, which is moot
now that #879 is merged; a correcting comment is posted on #879 pointing here.

---

# Round 2 — `a4c20c61f`

The round-1 adversarial review returned **0 BLOCKING, 4 SHOULD-FIX, 6 NOTE**. Every finding is
either fixed here or filed as #884 with its reason. The reviewer's own measurements were
re-derived rather than taken on report.

## SF-C — this PR had the defect it was fixing, one column up

The round-1 fix closed "the `T` half has no fixture" for the **worktree** column and left the
**index** column in exactly that state. The reviewer's mutant:

```
    && (gone(raw(path)[1]) || (raw(path)[1] === ' ' && gone(raw(path)[0]))));
->  && (gone(raw(path)[1]) || (raw(path)[1] === ' ' && raw(path)[0] === 'D')));

SURVIVED — pass=937 fail=0
```

The two-column fixture now stages the retype (`T `) and carries `MT` besides — the shape the #879
reviewer measured independently in its own lab at two shas. Five codes asserted by `deepEqual`,
not a count.

## SF-B — the ABSENT clause named the wrong store

`AD`/`AT` are staged additions removed from the worktree again. HEAD carries no such path, so for
them the pack *equals* the commit, and "the pack LACKS a file the commit has" is false — it is
the **index** that has it. That is the same shape of falsehood this function was fixed for
stamping on deletions in round 1.

They get their own clause:

```
REFUSED: 1 STAGED-BUT-GONE path(s) under a shipped directory so the pack lacks a file the
NEXT commit would have — `git restore <path>` brings it back, or unstage it:
  AD skills/real/added.md
```

## SF-D — the UNMERGED sentence described two of the seven codes it names

Measured per code: `UU` and `AA` pack conflict markers; `UA`/`AU`/`UD`/`DU` pack one side's
version with **no markers**; `DD` packs nothing at all. The verb "resolve it" is right for all
seven, so only the false half of the sentence goes.

## N1 — four of seven unmerged codes had no fixture

`UU`, `AA`, `UD` and `DU` each survived deletion from `UNMERGED_CODES` against all 937 tests: the
rename/rename conflict reaches only `DD`/`UA`/`AU`, while the comment said "every unmerged code".
A second fixture covers the other four, with modify/delete in **both** directions — `UD` needs
*our* side to modify, since a one-sided delete merges cleanly and reports `D `.

## N3, N5, N6

- **N3** — the fixture manifest had no `version`, so `npm pack --dry-run` exits 1 with
  `Invalid package, must have name and version`. The count in the "measured on this exact tree"
  line was right; the tree it was measured on was not the one the test builds. `version` added.
- **N5** — `["!lib/x/", "./lib"]`: npm normalises the prefix and packs `lib/a.js` and
  `lib/x/b.js`; this module compares literal paths, packs neither, and the negation goes dead.
  Refused rather than normalised — normalising means reimplementing npm's path handling on one
  measurement.
- **N6** — the ABSENT remedy names `git restore --staged --worktree` for a `D ` path, where plain
  `git restore` exits 1 with `pathspec … did not match any file(s) known to git`.

## Deliberately NOT fixed here: SF-A and N2 → #884

Both are the same root: this guard classifies from git's two-column code, and **npm reads the
disk**. `git rm --cached` emits two records for one path (`D ` + `??`) and draws two contradictory
false sentences — and that is the clean-up path for a mistakenly committed artefact under
`skills/`, the #872/#868 shape. `TM` and `TT` land in the wrong bucket for the same reason.

Left out because the routing **predates this PR** (`c52d074f1`'s guard prints the same two lines,
measured) and because swapping the classification source to `lstatSync` deserves its own
measurement and its own review round. What this PR does fix is the comment that asserted the
false premise behind it — a blank worktree column was documented as "the worktree agrees with the
index", and for `D ` the index has no entry left to agree with.

## Mutants — 12 rows, all under `npm run test:scripts`

| mutant | verdict |
|---|---|
| `sfc-absent-drops-T-index-column` | KILLED by 1 |
| `sfb-staged-gone-folds-into-absent` | KILLED by 1 |
| `sfd-unmerged-sentence-reverts` | KILLED by 1 |
| `n1-unmerged-drops-UU` | KILLED by 1 |
| `n1-unmerged-drops-AA` | KILLED by 1 |
| `n1-unmerged-drops-UD` | KILLED by 1 |
| `n1-unmerged-drops-DU` | KILLED by 1 |
| `n1-unmerged-drops-DD` | KILLED by 1 |
| `n5-dot-slash-refusal-disabled` | KILLED by 1 |
| `r1-worktree-column-reverts-to-trim` | KILLED by 1 |
| `r1-absent-drops-T-worktree-column` | KILLED by 1 |
| `r1-negation-prefix-loses-segment-boundary` | KILLED by 1 |

`0 not a clean kill`. The three round-1 rows are re-measured at this head rather than quoted from
the round-1 table.

## Gates at `a4c20c61f`

```
npm run test:scripts   → 938 pass, 0 fail, exit 0
npm test               → exit 0 (integrity, check-readmes, test:scripts, test:cli)
node scripts/check-publishable-tree.js → OK, exit 0
```

---

# Round 2 follow-up — `fdc46ecc2`

Two things the round-2 section above owed and did not carry. Neither changes the code under
test; both change what a reader can check.

## The twelve-row table had no committed plan

The rows were measured, but in a session scratchpad that no longer exists, so nothing committed
reproduced them — which makes a published count HISTORICAL by this repository's own convention,
and a reviewer right to refuse the table rather than grade it.

The plan is now a file and the table is its output:

```
bash tools/mutation-envelope.sh --test 'npm run test:scripts' \
  --plan tests/results/2026-09-22-publishable-tree-columns/mutation-plan.tsv
```

Re-measured at `29cfe074a` — twelve rows, every one `MUTANT KILLED by 1 failing test(s)`,
`0 not a clean kill`. The three `r1-` rows are round 1's claims re-taken at this head, not
carried over. The verdicts and the one qualification about how they were taken — a row costs
about 74 s, so the plan was run in three parts under the same `--test` command — are in
`tests/results/2026-09-22-publishable-tree-columns/RESULT.md`.

## N4 was not disposed of

Round 1 returned six notes. The round-2 section names N1, N2, N3, N5 and N6, and N4 appears
nowhere — which reads as dropped rather than deferred.

N4: the `assertInterpretable` leak fix is measured, not guarded. Removing its
`t.after(() => rmTree(dir))` survives the whole suite, because `tmp-helper.test.js` guards *how*
a fixture is removed and nothing guards *whether* one is removed at all.

Filed as **#885** rather than fixed here, for the two reasons this repository does not add a gate
in a bugfix PR: the accept rule is a design decision — `try { … } finally { rmTree(dir) }` is an
equally legitimate second shape, and is what `translation-status.test.js` and
`gate-envelope.test.js` actually use — and the member list needs reading before it is ratcheted.

**The member list is one file.** An earlier revision of this section published the one-shape
count — 20 of the 43 `mkdtempSync`-using suites — while calling it an over-count in the same
sentence, which gives a figure nobody has checked the authority of a finding. Measured instead,
by `tests/results/2026-09-22-publishable-tree-columns/teardown-predicate.sh`: counting BOTH
shapes leaves **one** suite, `memory-blocks.test.js`, and it leaves **21 directories in an
isolated `TMPDIR` per run** against 0 for the control suite. The refined predicate is still
crude — a `finally` that removes the wrong thing passes it — so it is one file to read, not a
ratchet. #885 is corrected to match.

---

# Round 3 — `60203b64f`

The round-2 adversarial review returned **0 BLOCKING, 3 SHOULD-FIX, 2 NOTE**. Every finding is
fixed here. Each of the three should-fixes was re-derived independently before being touched,
because a reviewer's measurement taken on report is not a measurement.

## SF-1 — the arm was pinned by a COUNT, and two mutants survived

The sharpest finding of the round, and it is the round-2 fix's own defect one level along. The
STAGED-BUT-GONE block was asserted by its header count. A count is preserved by a swap:

```
raw(path)[0] === 'A'  ->  raw(path)[0] === 'T'                         SURVIVED — 938 pass, 0 fail
raw(path)[0] === 'A'  ->  raw(path)[0] === 'A' && raw(path)[1] === 'D' SURVIVED — 938 pass, 0 fail
```

Both re-run here before the fix, both reproduce. The first puts `AD` back under
ABSENT-OR-RETYPED — "a file the commit has", the exact sentence SF-B was written to stop — and
moves `T ` into STAGED-BUT-GONE, leaving `4` and `1` where they were. The second narrows the arm
to `AD`, and `AT` slips out unnoticed because no fixture built one.

Fixed by asserting **membership**, not counts. A `refusedBlock(lines, label)` helper returns a
block's `<code> <path>` lines, and both blocks are compared with `deepEqual`; the fixture gains
an `AT` shape, so it now builds six codes rather than five. Both mutants are rows in the plan
and both are `KILLED by 1`. The test title, which still said "`AD` … are absences" after SF-B
had made that false, is corrected in the same edit.

## SF-2 — the ABSENT remedy was false for `T `, the code this PR added a fixture for

Round 1's N6 measured `D ` and said so; the fix generalised from one code. Measured here per
code, one fresh repository per (code, command) pair, on git 2.43.0 —
`tests/results/2026-09-22-publishable-tree-columns/restore-remedy-matrix.sh`:

| code | `git restore <path>` | `git restore --staged --worktree <path>` |
|---|---|---|
| ` D` | exit 0, clean | exit 0, clean |
| `D ` | exit 1, `pathspec … did not match` | exit 0, clean |
| ` T` | exit 0, clean | exit 0, clean |
| `T ` | exit 0, **NO CHANGE — still `T `** | exit 0, clean |
| `MD` | exit 0, leaves `M ` — the staged edit is recovered | exit 0, clean — **the staged edit is discarded** |
| `MT` | exit 0, leaves `M ` — the staged edit is recovered | exit 0, clean — **the staged edit is discarded** |

So the sentence was wrong in both directions: plain restore on `T ` is a silent exit-0 no-op, and
the two-flag form on `MD`/`MT` destroys work. The remedy now names a form per code, the matrix
sits in a comment beside it, and the assertion reads the clause rather than the block heading —
a revert of the parenthetical is a plan row and is `KILLED by 1`.

## SF-3 — a figure published under a predicate the same sentence called wrong

The N4 section quoted "20 of the 43 `mkdtempSync`-using suites" and called it an over-count in
the same breath. Counting **both** teardown shapes leaves one suite, and it does leak:

```
suites using mkdtempSync:                 43
crude   (mkdtempSync > t.after):          20
refined (mkdtempSync > t.after+finally):   1   memory-blocks.test.js
memory-blocks.test.js:     21 directories left in an isolated TMPDIR
publishable-tree.test.js:   0   (the control row)
```

The control row is load-bearing — a probe that counted nothing anywhere would otherwise read as
a clean result. The N4 section above is corrected, and #885 carries the same correction.

Running the whole suite rather than one file leaves **22** entries, and two of them need a
clause that a bare total invites dropping: `node-compile-cache` is **npm's**, not a fixture and
not the suite's — a bare `node --test` under a fresh `TMPDIR` leaves 0 and `npm run` of any
script leaves that one — and the `memcap-arms-` row is present **only with `python3` on
`PATH`**, since its test carries `skip: PYTHON3 === null` (`memory-blocks.test.js:277`).
Without `python3` the total is 21. Every leaked *fixture* still belongs to
`memory-blocks.test.js` (#883 round 3 delta, N-A and N-B).

## N-1, N-2

- **N-1** — the ordering comment claimed testing unmerged codes first stops the worktree arm
  claiming `DD`, `DU` or `UD`. `DU`'s worktree column is `U`, which `gone` never matches, so no
  ordering was protecting it. Dropped from the sentence.
- **N-2** — RESULT.md predicted that one envelope call reproduces the table. Nobody had run it
  that way. The prediction is gone; what is recorded is the reviewer's own measurement (the
  twelve-row plan in one call, exit 0, 53 s a row under the CI command) and the arithmetic that
  forced the split here.

## Mutants — 15 rows, all under `npm run test:scripts`

Every row re-taken at `342a94777`, three added this round:

```
sfc-absent-drops-T-index-column            KILLED by 1
sfb-staged-gone-folds-into-absent          KILLED by 1
sfd-unmerged-sentence-reverts              KILLED by 1
n1-unmerged-drops-UU                       KILLED by 1
n1-unmerged-drops-AA                       KILLED by 1
n1-unmerged-drops-UD                       KILLED by 1
n1-unmerged-drops-DU                       KILLED by 1
n1-unmerged-drops-DD                       KILLED by 1
n5-dot-slash-refusal-disabled              KILLED by 1
r1-worktree-column-reverts-to-trim         KILLED by 1
r1-absent-drops-T-worktree-column          KILLED by 1
r1-negation-prefix-loses-segment-boundary  KILLED by 1
sf1-staged-gone-A-to-T                     KILLED by 1   (SURVIVED last round)
sf1-staged-gone-narrows-to-AD              KILLED by 1   (SURVIVED last round)
sf2-absent-remedy-names-one-form           KILLED by 1

15 row(s), 0 not a clean kill
```

A row costs ~74 s here, so the plan is run in three parts of five under the same `--test`
command; what a split costs is one exit code over the set, which the three `0 not a clean kill`
lines stand in for.

## Gates at `60203b64f`

```
npm test                               → exit 0 (938 script pass, 141 CLI, 0 fail)
node scripts/check-publishable-tree.js → OK, exit 0
npm run validate:integrity             → exit 0
npm run check:tools-registry           → exit 0
npm run check:generated-artifacts      → exit 0
```

## Still deliberately out

#884 — this guard classifies from git's code table and npm reads the disk. Round 2 confirmed by
enumeration that `TT` lands in ABSENT-OR-RETYPED while the pack carries the file, and that the
`git rm --cached` routing predates this PR. Unchanged here; it needs `lstatSync` and its own
round.

---

# Round 4 — `1ff8bf508`

The round-3 adversarial review returned **0 BLOCKING, 2 SHOULD-FIX, 1 NOTE**. Both should-fixes
were re-derived here before being touched; both reproduced.

## SF-1 — the arm was pinned on the codes it must INCLUDE, not the one it must EXCLUDE

The same class a fourth time, one code along. Round 2 fixed a count that a swap preserved;
round 3 finds that membership over the included codes says nothing about an excluded one:

```
    && worktreeAbsent(path) && raw(path)[0] === 'A');  ->      && raw(path)[0] === 'A');
SURVIVED — 938 pass, 0 fail
```

On a tree with a staged add and a staged-then-edited add — the commonest divergence a
maintainer produces — the mutant files `A ` and `AM` under STAGED-BUT-GONE, "the pack lacks a
file the NEXT commit would have", and recommends `git restore <path>`, which is an exit-0 no-op
on `A `. The pack carries both files. Nothing caught it because no test read the *report* for
an `A ` path: the two-column fixture built none, the rename-WITHIN test asserts `divergentPaths`
output only, and rename-OUT asserts the path appears in any list.

The fixture now builds **eight** codes, and MODIFIED is asserted by membership like the other
two blocks. The mutant is a plan row and is `KILLED by 1`.

That collides with the comment below it, which claimed `npm pack --dry-run --json` listed
package.json alone on that tree. True before `A `/`AM` joined the fixture, false after — the
round-1 N3 class again. Rather than reword it from memory, the tree is rebuilt and npm is asked
by a committed script, `tests/results/2026-09-22-publishable-tree-columns/two-column-pack.sh`:

```
git reports:  A  new.md   AD added.md   AM new2.md   AT addsym.md
              MD SKILL.md  MT retyped.md  T  staged.md   T references/helper.py
npm packs:    package.json, skills/real/new.md, skills/real/new2.md   — 3 files
```

Exactly the two paths whose worktree file exists — which is the whole basis for splitting the
remedy sentences by bucket.

## SF-2 — a duration attributed to the wrong command

RESULT.md said the round-2 reviewer had run the twelve-row plan "in a single envelope call
under this file's own command". This file's command is `npm run test:scripts`. What was run in
one call was the **file-level** command, `node --test scripts/test/publishable-tree.test.js`:
12 KILLED, exit 0, **53 s in total**. Under `npm run test:scripts` a single **row** cost 53 s in
that same lab. The two figures coincide and mean different things, and merging them is exactly
what the plan file's own header warns about. No one-call run of this plan under the CI command
has been taken in any lab; the file now says so rather than implying otherwise.

## N-1 — the parser folded the truncation line into a block

`refusedBlock` took every following line starting with two spaces, and `  … and N more` starts
with two spaces. It failed closed — a `deepEqual` against real paths cannot pass — but it would
report a member that does not exist. The loop stops there, and the helper now has a unit test of
its own: a synthetic two-block report, an absent label returning `null`, and `report`'s own
truncation output over 21 paths, asserting the last member is a path.

## Mutants — 16 rows, all under `npm run test:scripts`

```
sfc-absent-drops-T-index-column            KILLED by 1
sfb-staged-gone-folds-into-absent          KILLED by 2
sfd-unmerged-sentence-reverts              KILLED by 1
n1-unmerged-drops-UU                       KILLED by 1
n1-unmerged-drops-AA                       KILLED by 1
n1-unmerged-drops-UD                       KILLED by 1
n1-unmerged-drops-DU                       KILLED by 1
n1-unmerged-drops-DD                       KILLED by 1
n5-dot-slash-refusal-disabled              KILLED by 1
r1-worktree-column-reverts-to-trim         KILLED by 2
r1-absent-drops-T-worktree-column          KILLED by 1
r1-negation-prefix-loses-segment-boundary  KILLED by 1
sf1-staged-gone-A-to-T                     KILLED by 2   (SURVIVED at 12 rows)
sf1-staged-gone-narrows-to-AD              KILLED by 1   (SURVIVED at 12 rows)
sf1r3-staged-gone-drops-worktree-test      KILLED by 1   (SURVIVED at 15 rows)
sf2-absent-remedy-names-one-form           KILLED by 1

16 row(s), 0 not a clean kill
```

Three rows kill by 2, and that is not #621's crash signature: the second failing test is the new
`refusedBlock` unit test, which calls `report` directly, so a mutation of the bucket predicates
is asserted twice — once through the fixture, once through the parser's test. Both assertions
are about the property.

## Gates at `1ff8bf508`

```
npm test                               → exit 0
node scripts/check-publishable-tree.js → OK, exit 0
npm run validate:integrity             → exit 0
npm run check-readmes                  → exit 0
npm run check:tools-registry           → exit 0
```

## Recorded, not fixed

Round 3 measured that dropping the blank-worktree conjunct — `|| (raw(path)[1] === ' ' &&
gone(raw(path)[0]))` → `|| gone(raw(path)[0])` — survives at file level, and that the only
porcelain code it moves is `TM` (MODIFIED → ABSENT-OR-RETYPED). `TM`'s pack is a dropped
symlink, so the mutant is arguably *more* right about it, and which bucket `TM` belongs in is
#884's question. An untested conjunct whose only reachable effect is on a deferred code.

---

# Round 5 — `d0e91c97e`

The round-4 review returned **0 BLOCKING, 1 SHOULD-FIX, 3 NOTE**. All four are applied, and the
should-fix had already been found and fixed on this branch independently, by the same remedy the
reviewer names.

## SF-1 — "KILLED by 2" was one property kill plus a crash

The round-4 body published this sentence: *"Three rows kill by 2 … the second failing test is
the new `refusedBlock` unit test … Both assertions are about the property."* It was written from
the shape of the tests, not from their output. Measured:

```
sfb-staged-gone-folds-into-absent   two-column test: AssertionError deep-equal
                                    parser test:     TypeError: Cannot read properties of null (reading 'length')
```

The parser test hardcoded `refusedBlock(truncated, 'STAGED-BUT-GONE')`. Any mutant that routed
`AD` elsewhere made the block absent, `refusedBlock` returned `null`, and `.length` threw — a
crash in a test asserting nothing about the mutated line. That is #621's shape with the roles
reversed, and it is the fourth instance of the class this PR has been closing one step at a
time.

Fixed by making the test independent of routing rather than by rewording the claim: it reads the
label back out of `report`'s own first `REFUSED:` line, with two `assert.ok` guards so a null
fails with a sentence instead of a dereference. All three rows are re-measured at
**`KILLED by 1`**.

Worth recording, because it decides how such a count should be read: **the checker could not
have caught this.** Its broad-kill heuristic trips at `BROAD_KILL_SHARE = 0.25` of the baseline,
and 2/939 is 0.002 — the count was never going to be questioned. A `by 2` needs a human to read
which two.

## N-3 — the parser's own fixture could not tell `slice(2)` from `trim()`

The synthetic block held only two-character codes, over which the two are indistinguishable.
Measured: mutating `out.push(line.slice(2))` to `line.trim()` left the parser test passing
(1 pass), and it was killed only by the fixture test's ` T` member. A space-leading member
(`   T skills/real/retyped.md`) is in the synthetic block now; the same mutant then fails the
parser test directly (1 fail).

## N-1 — the pack script duplicated the fixture with nothing tying the two

`two-column-pack.sh` rebuilds the test's fixture by hand, so a fixture change would leave it
reporting the same three files beside a comment that had gone false again — with a green script
next to it. The eight porcelain codes are now declared in the script and compared, so drift
**refuses** rather than reporting. Exercised by renaming one expected path:

```
REFUSED: this fixture no longer builds the eight codes the two-column test asserts.
4c4
< AM skills/real/new3.md
---
> AM skills/real/new2.md
```

Its header also claimed it writes only under its own `mktemp -d`. False by one entry: the
`npm pack` it runs makes npm write a `node-compile-cache/` into `TMPDIR`, outside `DIR`, which
the trap does not remove. Corrected.

## N-2 — a stale count in prose

RESULT.md said "the twelve mutants" over a sixteen-row plan. The count now lives in the plan and
in the envelope's own summary line.

## Mutants — 16 rows, every one `KILLED by 1`

The three former by-2 rows re-measured after the parser fix:

```
sfb-staged-gone-folds-into-absent   KILLED by 1
r1-worktree-column-reverts-to-trim  KILLED by 1
sf1-staged-gone-A-to-T              KILLED by 1
```

One failing test per row, which is the honest shape for this instrument — and, as this round
showed, a kill by two can hide a crash just as a broad kill can.

## Gates at `d0e91c97e`

```
npm test                               → exit 0 (939 script, 141 CLI, 0 fail)
node scripts/check-publishable-tree.js → OK, exit 0
npm run validate:integrity             → exit 0
npm run check-readmes                  → exit 0
npm run check:tools-registry           → exit 0
```

---

# Round 6 — `a17fa0f1c`

The round-5 review returned **0 BLOCKING, 1 SHOULD-FIX, 1 NOTE**. It found the fifth instance of
the class, which is what it was asked to look for, and it is the cleanest one of the six rounds.

## SF-1 — the drift guard compared the script to itself

`two-column-pack.sh` named `scripts/test/publishable-tree.test.js` in three comments and one
error message, **read nothing from it**, and declared a literal of its own. So the only drift it
could see was between two things in the same file, while its comment — and the round-5 section
of this body — said it refused when the *test's* fixture changed.

Measured, in a lab built from the object store: rename `new2.md` to `new3.md` at all six sites
of the **test's** fixture, and

```
node --test scripts/test/publishable-tree.test.js   → 27/27 pass
bash …/two-column-pack.sh                           → exit 0, zero REFUSED lines,
                                                       still reporting new2.md
```

A check described as guarding X that never reads X. The exercise this body offered as evidence
for the guard had renamed the script's **own literal**, which is the one direction a
self-comparison does cover — so the demonstration and the claim were about different things.

Fixed by removing the second source rather than by re-syncing it. `expected-codes.mjs` parses
the test's `deepEqual` on `found.codes` and prints the eight codes; the script calls it and
compares its fixture against that. There is one source now, so there is nothing left to drift.
Three directions measured:

```
test fixture renamed, script not   exit 1, REFUSED, the diff names AM new3.md vs new2.md
assertion message reworded         exit 2, "would compare its fixture against nothing and pass"
neither                            exit 0, 3 files packed
```

The matcher is anchored on the assertion message rather than on position, and a failed match
exits 2 with **nothing on stdout** — a caller that captured an empty success would compare its
fixture against nothing and pass, which is the failure mode this whole class keeps producing.

`check-drift-literal.mjs`, added earlier in this round to check one literal against the other,
is removed: there is only one literal now.

## N-1 — "at `e0df06cb3` and after" is not a provenance

Corrected to name the three shas that produced the table: the sixteen rows taken as a set at
`d19981306`, the three that read `by 2` re-taken after `e0df06cb3`, and the reviewer's own
one-call run of all sixteen at `d0e91c97e` reading every row `by 1`.

## Also fixed, from round 5's "what holds"

`NODE_DISABLE_COMPILE_CACHE=1` in the pack script. Round 4 documented that the `npm pack` it
runs makes npm write a `node-compile-cache/` outside the script's own `mktemp -d`; round 5
measured that the env var stops the write rather than relocating it, and that the script still
reports 3 files. So the header's claim to write only under its own directory is now true instead
of caveated — measured, nothing left outside `DIR` under an outer `TMPDIR` sandbox.

## Gates at `a17fa0f1c`

```
npm test                               → exit 0 (939 script, 141 CLI, 0 fail)
node scripts/check-publishable-tree.js → OK, exit 0
npm run validate:integrity             → exit 0
npm run check-readmes                  → exit 0
npm run check:tools-registry           → exit 0
npm run check:generated-artifacts      → exit 0
```

## The class, across six rounds

Each round found the same shape one step along, and naming them together is the point:

| round | what passed for the wrong reason |
|---|---|
| 1 | the `T` half of a predicate had no fixture |
| 2 | the block was pinned by a **count**, which a swap preserves |
| 3 | membership over the codes an arm **includes**, saying nothing about the one it excludes |
| 4 | a **crash** in an unrelated test counted as the second kill of a `by 2` |
| 5 | a guard comparing a thing **to itself** while claiming to guard another |

The instrument was honest in every round; what was wrong each time was a claim about what the
instrument covered.

---

# Round 7 — `aaba16c14`

The round-6 review returned **0 BLOCKING, 2 SHOULD-FIX, 1 NOTE**, and answered the question this
PR actually turned on:

> **Q4 first, since it is the question that matters. Yes.** The guard has not changed since
> `342a94777`; the test file since `d0e91c97e`; all sixteen plan rows are `KILLED by 1` at that
> code; `npm run test:scripts` is 939 green. […] Both findings below live in
> `tests/results/2026-09-22-publishable-tree-columns/` and neither touches what ships.

Both are fixed here anyway. Neither is in `scripts/`.

## SF-1 — the delta's finding, carried into the successor

The round-5 delta review closed with *"a parser that drops what it cannot read is the same
instrument whichever file calls it"*. It was right. `expected-codes.mjs` kept the predecessor's
regex, and a ninth entry the regex could not see contributed nothing to either side:

```
double-quoted ninth entry in the test's deepEqual
  expected-codes.mjs  → 8 lines, exit 0
  two-column-pack.sh  → exit 0, zero REFUSED lines     (the test asserted NINE)
```

Fixed at `b0019bbc9`, before the round-6 report arrived — every non-blank line inside the block
parses as an entry or is a comment, anything else exits 2; the anchor moved to the test's title;
the extent is taken by brace matching. The reviewer's second shape, a **nested-object** ninth
entry, is confirmed refused at exit 2 here.

`prove-expected-codes.sh` is the negative evidence — six arms, one `git archive` lab each, three
of which must NOT refuse, because a rule that refused everything would pass a matrix made only
of bad rows.

## SF-2 — the rewiring compared two different sort orders

New this round, and invisible to both labs because both run under a C-family locale. The
round-5 script piped its own literal through the same `sort` as the porcelain, so both sides
collated alike. Reading the codes from the test made `EXPECTED` come from JavaScript's
`Array.prototype.sort` — code-unit order — while `PORCELAIN` still went through coreutils
`sort`, which collates by locale. Under `en_US.UTF-8`, leading whitespace is ignored at the
first collation level, so `A  new.md` sorts after `AD added.md`.

Reproduced here with a privately compiled locale (`localedef -i en_US -f UTF-8`), fixed with
`LC_ALL=C sort`:

```
pre   C            exit=0  clean
pre   en_US.UTF-8  exit=1  REFUSED
post  C            exit=0  clean
post  en_US.UTF-8  exit=0  clean
```

It failed closed — a false refusal in a record script, never a false pass.

## N-1 — a code does not determine packedness

The codes tie is sufficient only under an assumption neither fixture states. Measured by the
reviewer: ` T` **packs** the file when the committed object was a symlink retyped to a regular
file, and **drops** it in the fixture's own direction. "Same codes" implies "same pack" only
because both fixtures commit regular files, and nothing checks that. The assumption is now
written where the comparison happens, with the durable form named for if it ever bites.

## Q3 — the instrument row, added with its scope sentence

`trim-eats-the-space-leading-code` joins the plan under a labelled **INSTRUMENT ROWS** section.
It mutates the test's own helper rather than the subject, and it is the only row that reads
`KILLED by 2` — which round 4 established must be *read*, never trusted:

```
✖ refusedBlock reads one block, and not the truncation line
  AssertionError [ERR_ASSERTION]: the space-leading code survives the slice — `trim()` would silently eat it
✖ a TWO-COLUMN code is read at the WORKTREE column, and an `A?` path is not one the commit has
  AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal
```

Both `AssertionError`, both named, neither a crash. The scope sentence, in the plan: a parser
defect can only redden the baseline, it cannot make a **subject** row survive — so the other
sixteen remain claims about `check-publishable-tree.js` and a reader counting them that way is
still right.

## Mutants — 17 rows, 16 subject + 1 instrument

```
16 subject rows                      KILLED by 1
trim-eats-the-space-leading-code     KILLED by 2   [instrument, both AssertionError]

17 row(s), 0 not a clean kill
```

## Gates at `aaba16c14`

```
npm test                               → exit 0 (939 script, 141 CLI, 0 fail)
node scripts/check-publishable-tree.js → OK, exit 0
npm run validate:integrity             → exit 0
npm run check-readmes                  → exit 0
npm run check:tools-registry           → exit 0
npm run check:generated-artifacts      → exit 0
bash …/prove-expected-codes.sh         → exit 0, six arms
bash …/two-column-pack.sh              → exit 0, 3 files
```

## The one design decision left open

Nothing *runs* the three probes in `tests/results/` — they are invoked by hand. The reviewer's
named exit is to retire `two-column-pack.sh` into an `npm pack --dry-run --json` assertion
inside the test itself. I have not taken it: that adds a second-per-run subprocess to a required
CI context for a record-keeping property. Stated here rather than settled silently, so the next
session can take it if the trade looks different to them.

---

# Round 8 — `0ad5e65e9`

Round 7 came back **clean — `BLOCKING=0 SHOULD-FIX=0 NOTE=0`, "Merge on it."** Two commits
land after the sha it graded, both in `tests/results/`, neither touching what ships.

## The three delta NOTEs from round 6, fixed at `433fa747c`

- **N-1** — the character walk counted a `}` inside a comment as a closer, ending the block
  early and silently; the truncated set compared equal and the guard passed. Extent and
  accounting are now one line-wise pass with comment lines skipped **before** counting.
- **N-2** — the anchor's claim (a reworded message cannot move it onto another block) held only
  while the title prefix was unique. A duplicated prefix now refuses.
- **N-3** — each lab is `git archive HEAD` overlaid with working-tree copies, so a proof of
  uncommitted work looked identical to a proof of the committed state. The transcript prints the
  sha and whether those copies are dirty.

Two arms join the matrix, now eight. The `}`-in-a-comment arm places its comment after the
**first** entry, not the last, and that placement *is* the arm: after the last key a premature
close truncates nothing and the arm would pass against the very parser it exists to catch.
Measured against the pre-fix parser — 1 entry and a false refusal; after — 8 and exit 0.

## Round 7's one nit, fixed at `0ad5e65e9`

The comment beside the `LC_ALL=C` fix said "only a byte order is common to both". It claimed
more than the code guarantees, which is the shape this whole PR has been about. `LC_ALL=C sort`
orders UTF-8 **bytes**; JavaScript's `.sort()` orders UTF-16 **code units**. Measured:

```
JS sort:  ["𐀀x", "～y"]
LC_ALL=C: ～y  then  𐀀x
```

They agree on ASCII and most of the BMP — every path any fixture here has had — and a
disagreement would be a spurious refusal rather than a silent pass. The comment now names which
order each side uses and the remedy for a fixture that ever needs an astral path.

## Eight rounds, seven instances of one class

| round | what passed for the wrong reason |
|---|---|
| 1 | the `T` half of a predicate had no fixture |
| 2 | the block was pinned by a **count**, which a swap preserves |
| 3 | membership over the codes an arm **includes**, saying nothing about the one it excludes |
| 4 | a **crash** in an unrelated test counted as the second kill of a `by 2` |
| 5 | a guard comparing a thing **to itself** while claiming to guard another |
| 6 | a parser **dropping** what it cannot read, so an addition reads as identity |
| 7 | a comment **claiming more** than the code guarantees |

Every instrument was honest in every round. What was wrong each time was a claim about what the
instrument covered — which is why the reviewer's measurements, not its opinions, are what moved
this PR.

## Gates at `0ad5e65e9`

```
npm test                               → exit 0 (939 script, 141 CLI, 0 fail)
node scripts/check-publishable-tree.js → OK, exit 0
npm run validate:integrity             → exit 0
npm run check-readmes                  → exit 0
bash …/prove-expected-codes.sh         → exit 0, eight arms
bash …/two-column-pack.sh              → exit 0, 3 files
```

`scripts/check-publishable-tree.js`, `scripts/lib/skills-inventory.js` and
`scripts/test/publishable-tree.test.js` are byte-identical to `d0e91c97e`, the sha round 6
called mergeable.


pjt222:
Merged as `d6b9b9c72dea1f88ab60f84c7611dfedc16790d4` after **nine adversarial rounds** (advocatus-diaboli on Fable, each round graded against a bundle cut at a named sha). Final round: `BLOCKING=0 SHOULD-FIX=0 NOTE=0`.

Follow-ups, all filed:

- **#884** — the classification source. This guard reads git's two-column porcelain code; npm reads the disk. They disagree for `git rm --cached` (two records for one path) and for `TM`/`TT`. Pre-existing, measured, deliberately outside this PR.
- **#885** — the fixture-teardown gate. `tmp-helper.test.js` guards *how* a fixture is removed and nothing guards *whether* one is. Member list measured end to end: exactly one suite.
- **#886** — the reviewer's own recommendation, recorded as theirs with its discriminating fact: **the guard's premise — npm packs an on-disk regular file and drops a symlink — is asserted nowhere in CI today.** One `npm pack --dry-run --json` assertion in the two-column test would assert it and retire four record probes. The trade was weighed and not taken here (a subprocess in a required CI context for a record-keeping property); the issue exists so a later session can take the other side with the measurement in hand.

## What the nine rounds were actually about

The two-column read was correct by round 3. Everything after that was one recurring class, found one step along each time:

| round | what passed for the wrong reason |
|---|---|
| 1 | the `T` half of a predicate had no fixture |
| 2 | the block was pinned by a **count**, which a swap preserves |
| 3 | membership over the codes an arm **includes**, saying nothing about the one it excludes |
| 4 | a **crash** in an unrelated test counted as the second kill of a `by 2` |
| 5 | a guard comparing a thing **to itself** while claiming to guard another |
| 6 | a parser **dropping** what it cannot read, so an addition reads as identity |
| 7 | a comment **claiming more** than the code guarantees |
| 8 | the fix for 6 **regressed** on a shape its predecessor refused |

Every instrument was honest in every round. What was wrong each time was a claim about what the instrument covered — twice in a fix written for the previous instance of exactly that.

Two figures published in this PR were false when written and corrected by measurement rather than by re-reading: "both assertions are about the property" (one was a `TypeError`) and "20 of the 43 suites" (the real member list is one file). Both corrections are in the body and, for the second, on #885.

Issue #901, opened 2026-09-23
[LOW] redact-visualization-for-disclosure Step 3: `attr:data-id[0]` is a label the tool cannot emit, and the exit-2 list omits the markup refusal


pjt222:
## Summary

`skills/redact-visualization-for-disclosure/SKILL.md` says three things about `tools/redact-artifact.py` that are no longer true at main `742a09271`:

1. **Step 3, :111** gives `attr:data-id[0]` as an example position label. The tool numbers positions from 1, so it cannot emit `[0]`. On the matching fixture it reports `attr:data-id[1]`.
2. **Step 3, :115-116** gives the causes of exit 2 as "an empty mapping, an unreadable input". The markup refusal is missing, and it is the one exit-2 cause specific to the `--type html` run that :109-111 recommends.
3. **Step 5, :145-150** reports every exit 2 as a leak. The `||` branch prints "rendered image still leaks" on any non-zero exit, and :150 still describes exit 2 with the wording #869 replaced at :115.

No gate catches any of them. The skill has no i18n mirrors, so the fix touches one file. This issue is related to #870 (same tool) but does not duplicate it; #870's body never mentions this skill.

All measurements below ran on a scratch copy of the two tool files in a `mktemp -d` directory, on Python 3.12.3, with `PYTHONDONTWRITEBYTECODE=1`:

```sh
DIR="$(mktemp -d)" || exit 1
cp tools/redact-artifact.py tools/redaction-lib.py "${DIR:?}/"
printf 'acme_secret\t[secret]\n' > "${DIR:?}/map.tsv"
```

## 1. `attr:data-id[0]` at :111

`rg -n 'attr:data-id' skills/redact-visualization-for-disclosure/SKILL.md`:

```
111:an attribute value, reporting the position (`text[2]`, `attr:data-id[0]`) and never the content.
```

Every `_Positions` handler increments the shared counter before it appends a label. `rg -n 'self\._n \+= 1' tools/redact-artifact.py` finds this at :142, :150, :155, :159 and :166, so the smallest ordinal an HTML label can carry is 1. This fixture produces both labels the skill quotes:

```sh
printf '<div data-id="acme&#95;secret">acme&#95;secret</div>\n' > "${DIR:?}/leak.html"
[...]

PR #909, opened 2026-09-23
fix(tools): redact-artifact refuses only input its parser dropped, and reports a term split across a block boundary (#870)

### LINKED ISSUE #870: [MED] redact-artifact: the structure-tier refusal fires on well-formed documents, and the BLOCK flush is fail-open
Two findings from adversarial round 3 against the merged #867. Both are in the structure tier; both were marked *issue* rather than *fix-commit* by the reviewer because each is a design change rather than a patch. Neither blocks anything shipped — round 3 returned `GATE: SHOULD-FIX — 0 blocking` and confirmed 8 of 8 prior blockers closed.

## 1. [MED] The refusal cannot tell an empty-but-valid document from a consumed one

`redact_text` refuses when `positions(out, kind)` is empty over non-empty input, on the theory that a structure tier which examined nothing must not report clean. The predicate is right; its implementation confuses two different things.

`not positions(...)` is true for any document with no text, no comments, no PIs or declarations, and no non-empty attribute value — **a normal class of valid HTML and SVG**:

```
  an empty div                                  REFUSED-AS-UNPARSEABLE
  a lone br                                     REFUSED-AS-UNPARSEABLE
  a lone hr                                     REFUSED-AS-UNPARSEABLE
  an img with no attributes                     REFUSED-AS-UNPARSEABLE
  an empty paragraph                            REFUSED-AS-UNPARSEABLE
  an SVG of shape elements with no attributes   REFUSED-AS-UNPARSEABLE
  an empty table                                REFUSED-AS-UNPARSEABLE
  an empty list                                 REFUSED-AS-UNPARSEABLE
  empty nested elements                         REFUSED-AS-UNPARSEABLE
  a real minimal SVG (control)                  clean
  ordinary prose (control)                      clean
  a document of one comment (control)           clean
  -> 9 of 12 well-formed documents refused
```

The denominator is not a random sample — those twelve were chosen to probe this boundary and the three controls show the predicate is not simply broken. `<div></div>` is reported with the words *"the document did not parse"*, which is false.

**`HTMLParser` can answer the real question**, measured:

```
  VALID but positionless    rawdata after feed=''                              after close='' positions=0
  MALFORMED (unterminated)  rawdata after feed='<text data-id="acme&#95;…'     after close='' positions=0
  valid with text           rawdata after feed=''                              after close='' positions=1
```

`rawdata` after `feed()` holds the unconsumed tail when the parser stalls and is empty when it did not — **but only before `close()`**, which flushes it either way. So the check belongs inside `_Positions`, taken between `feed` and `close` and exposed as a flag.

Two smaller facts in the same area: the `out.strip()` half of the predicate is unpinned (`malformed-refusal-drops-nonempty-guard` survives at 46/46), and the refusal **can never fire for `--type mermaid`** because `positions()` appends `document` unconditionally — yet the message interpolates `--type {kind}` as though it were general. Only `html` can reach it.

### Acceptance criteria

- [ ] `_Positions` exposes whether the parser stalled, captured between `feed()` and `close()`, and the refusal uses that rather than an empty position list
- [ ] A well-formed positionless document (`<div></div>`, `<br/>`, an attribute-free SVG) does **not** refuse
- [ ] A genuinely malformed document still refuses with exit 2
- [ ] Both arms are mutation-proven, and the `out.strip()` guard gains an arm
- [ ] The message no longer claims "did not parse" for a document that parsed, and no longer implies generality across types the refusal cannot reach

## 2. [MED] The `BLOCK` flush is fail-open, and round 3 is worse than round 2 here

Round 2 found that joining text runs across `</td><td>` produced an unactionable false positive. The round-2→3 fix flushes the run on 37 block-level tags. That traded a false positive for a **false CLEAN**, which is the harmful direction:

```
  two inline-styled DIVs render as one word     CLEAN
  two inline-styled Ps                          CLEAN
  inline-styled table cells                     CLEAN
  class-driven inline (CSS in a stylesheet)     CLEAN
  two SPANs (not in BLOCK) — the control        RAISED structure: term(s) survive at text[2]:acme_secret
```

and in the entity-encoded form, where tier 2 is the only tier that could see it at all:

```
  inline-styled divs: CLEAN
  spans (control)   : RAISED
```

The class docstring states the rule as *"a run of text is broken only by something a RENDERER breaks it with"* — and **that is not decidable from markup**. Whether a `<div>` breaks a run is a CSS `display` question and the tool never sees the stylesheet. The comment beside `BLOCK` correctly says *"erring toward flushing is fail-open … and only the second loses a leak"*, and the code then errs toward flushing on all 37 (`BLOCK members: 37; flush-and-lose on: 37`).

Severity stated plainly rather than inflated: this needs block-level elements with inline display carrying a *split* identifier. Real in generated markup, not common.

**The repair that keeps both properties**: do not drop a cross-block join — report it under its own label (`text-across-block[N]`), so a block-split term becomes a distinct, lower-confidence hit an operator can triage instead of a silence.

Second instance, **inferred from SVG semantics and not run**: `br` and `hr` are in `BLOCK` but are not rendered elements in SVG, so `<text>acme_<br/>secret</text>` (measured CLEAN) would render as one identifier. Treat as a hypothesis to check with a real renderer.

### Acceptance criteria

- [ ] A term split across a block boundary is reported under its own label rather than silently dropped
- [ ] The docstring stops claiming a renderer-accurate rule the tool cannot implement, and says what it actually does
- [ ] An arm covers the inline-styled-div case in both the plain and entity-encoded forms
- [ ] The SVG `br`/`hr` hypothesis is either measured against a renderer or recorded as unmeasured

## Also from the same round, low priority

`decl[]` and `pi[]` positions are **non-additive** — neither value is ever unescaped, so both are always substrings of the input the whole-text tier already checked. That is round-1 B2's class at small scale. Measured over three documents, zero additive entries; `no-decl-pi-positions` survives at 46/46. They are not harmful — they inflate the `N decoded position(s)` count and nothing else. Either `unescape` them, which would earn their place, or drop them. The run-joining behaviour they participate in is separately covered by the comment and PI arms, so removing the *positions* changes nothing.

## Context

Round 3 on #867, after merge. Reproduction scripts referenced in that round: `probes4.py` (12 probe sections) and `mutants3.py` (17 mutants). The fix-commit half of that round shipped as #869.

pjt222:
## Summary

Closes #870. Both of its findings are in `tools/redact-artifact.py`'s structure tier, and so is its low-priority item.

- **The refusal fired on well-formed documents and missed malformed ones.** It keyed on an empty position list. That is true of valid markup with no text: `<div></div>`, `<br/>` and an attribute-free SVG all exited 2 with "yielded no positions" (9 of the issue's 12 probes). It is false as soon as any text precedes a stall. The 2026-09-23 status comment measured `<p>ok</p>` followed by an unterminated attribute exiting 0, with the tail unexamined.
  - The refusal now keys on input the parser **dropped**: a tail `feed()` left unconsumed, for which `close()` then calls none of this class's handlers. That includes `unknown_decl`, where a CDATA section arrives. Uncounted, an unclosed `<![CDATA[` refused while a closed one passed, over a raw substring the whole-text tier had already checked.
  - A tail alone is not enough, as the comment warned. `AT&T` leaves one, a charref held back in case more input completes it, and `close()` emits it.
  - The refusal names `--type html` only, the one type that can reach it; an arm pins that. (The "did not parse" wording itself was already replaced by #869.)
  - A leak the tier finds is exit 1 even when the parser also dropped input after it. The message adds that the list may be incomplete. Only a clean result over a dropped tail refuses (round 1, below).
- **A term split across a block boundary was a false CLEAN.** A block tag flushed the text run and the join was dropped, so two `display:inline` divs rendering `acme_secret` as one word exited 0.
  - The split is now reported under its own label, `text-across-block[N]`, and it **exits 1 like any other survivor**. That is the maintainer's decision on this PR, taken over a separate triage exit code and over a warning at exit 0.
  - It includes table cells, which a renderer does separate. The self-test arm that called that join "an unactionable false positive" now asserts the label instead, since no override flag exists.
  - The class docstring stops claiming a renderer's rule, which depends on CSS the tool never sees, and says what markup can decide.
- **`decl[]` and `pi[]` positions are gone** (the issue's low-priority item). Neither is ever unescaped, so each value was a substring the whole-text tier had already checked. They inflated `N decoded position(s)` and could report nothing. The handlers still count, so no other label's ordinal moves.

## How the drop signal was chosen

Measured on Python 3.12.3 (`libpython3.12-stdlib 3.12.3-1ubuntu0.17`). The `tools-verify` job installs 3.12 through `actions/setup-python` (`validate-tools.yml`). The support session's claim check read the actions manifest's newest 3.12 as 3.12.14 on 2026-09-23. Upstream `v3.12.14:Lib/html/parser.py` carries the same `_pending` buffering with `_parse_threshold = 1`, while 3.12.3–3.12.13 upstream do not. That was read from source. The run then confirmed it: `tools-verify`'s log at `cfcea15dc` says `Successfully set up CPython (3.12.14)`, and the job passed, stall arms included. For each shape, the probe recorded the tail `feed()` left and how many handlers `HTMLParser.close()` itself called.

| input | tail after `feed()` | handlers during `close()` | examined? |
|---|---|---|---|
| `AT&T`, `fish & chips &`, `x &#95` | the text | 1 (data) | yes |
| `x <` | `<` | 1 (data) | yes |
| an unclosed comment / script / decl / PI | the construct | 1 | yes |
| an unclosed CDATA section | the construct | 1 (`unknown_decl`) | yes, once this class counts `unknown_decl` |
| an unterminated attribute (alone, or after `<p>ok</p>` or `x`) | `<text data-id="…` | 0 | **no, dropped** |
| an unclosed start tag or end tag | the construct | 0 | **no, dropped** |

A `<`-prefix rule would have refused `x <` and the unclosed comment, which is the same false refusal again. "Tail present and none of this class's handlers fired" separates the rows. "This class's" is load-bearing: the first draft did not count `unknown_decl`, so an unclosed CDATA section read as a drop. The claim check found that with a parser implementing every handler.

This local 3.12.3 carries a backported `HTMLParser` patch that buffers `feed()` input in `_pending`. Its initial `_parse_threshold` is 1, so the single `feed()` this tool makes parses at once and leaves the tail in `rawdata`. A parser that parsed nothing until `close()` would leave no tail, and the self-test's six stall arms would go red. So it is pinned, not assumed.

## Before and after

Exit codes through the CLI (`--type html --mapping {acme_secret: …} --also-deny acme_secret`). Before is `main` at `8a8355e01`, after is `cd934bec3`. The two `r1:` rows are round 1's SF-1 shapes. The cells show the exit code and the leading labels only. On this branch those two messages end "; the parser also dropped input it could not tokenise, so this list may be incomplete", and at `cfcea15dc` they exited 2. It covers the issue's twelve probes, the status comment's rows, and finding 2's rows. The doubled `text[1]` label comes from the term being passed as both a mapping key and an `--also-deny` term; it is not something this PR changes.

| input | before | after |
|---|---|---|
| issue: an empty div | 2 REFUSED | 0 clean |
| issue: a lone br | 2 REFUSED | 0 clean |
| issue: a lone hr | 2 REFUSED | 0 clean |
| issue: an img with no attributes | 2 REFUSED | 0 clean |
| issue: an empty paragraph | 2 REFUSED | 0 clean |
| issue: an SVG of attribute-free shapes | 2 REFUSED | 0 clean |
| issue: an empty table | 2 REFUSED | 0 clean |
| issue: an empty list | 2 REFUSED | 0 clean |
| issue: empty nested elements | 2 REFUSED | 0 clean |
| issue: a real minimal SVG (control) | 0 clean | 0 clean |
| issue: ordinary prose (control) | 0 clean | 0 clean |
| issue: one comment (control) | 0 clean | 0 clean |
| comment: malformed tail alone (control) | 2 REFUSED | 2 REFUSED |
| comment: `<p>ok</p>` + malformed tail | 0 clean | 2 REFUSED |
| comment: x + malformed tail | 0 clean | 2 REFUSED |
| comment: known leak (control) | 1 FAILED text[1]:acme_secret, text[1]:acme_secret | 1 FAILED text[1]:acme_secret, text[1]:acme_secret |
| f2: inline-styled divs | 0 clean | 1 FAILED text-across-block[2]:acme_secret |
| f2: inline-styled divs, &#95; | 0 clean | 1 FAILED text-across-block[2]:acme_secret |
| f2: inline-styled table cells | 0 clean | 1 FAILED text-across-block[2]:acme_secret |
| f2: two spans (control) | 1 FAILED text[1]:acme_secret, text[1]:acme_secret | 1 FAILED text[1]:acme_secret, text[1]:acme_secret |
| f2: svg text split by br (hypothesis) | 0 clean | 1 FAILED text-across-block[1]:acme_secret |
| r1: known leak + malformed tail | 1 FAILED text[1]:acme_secret, text[1]:acme_secret | 1 FAILED text[1]:acme_secret, text[1]:acme_secret (+ "may be incomplete" note) |
| r1: known leak + unclosed start tag | 1 FAILED text[1]:acme_secret, text[1]:acme_secret | 1 FAILED text[1]:acme_secret, text[1]:acme_secret (+ "may be incomplete" note) |

The last row is the issue's SVG `<br>` hypothesis. It is still **not measured against a renderer**, and the docstring says so. It now raises under `text-across-block` either way.

## Evidence

- **Self-test:** `python3 tools/redact-artifact.py --verify`: 79/79 arms (46 on `main`). No `__pycache__` is written under `tools/` (`PYTHONDONTWRITEBYTECODE=1`); `check:tools-registry` would reject one.
- **New arms:**
  - the nine positionless documents, plus empty and whitespace-only input, do not refuse;
  - six tails `close()` examines do not refuse: `AT&T`, a trailing `<`, an unclosed comment, CDATA section, doctype and XML declaration;
  - a leak found ahead of a dropped tail is exit 1 and says the list may be incomplete;
  - a mermaid label holding `<` is not refused, since the refusal is html-only;
  - a stall refuses after six different prefixes;
  - inline-styled divs, plain and entity-encoded, raise `text-across-block`, with a span control raising `text[`;
  - class-driven inline paragraphs and table cells raise it too;
  - a leak inside one run gets one label;
  - whitespace between blocks stays in the join;
  - the label names the run the term starts in, once per run.
- **Self-test hardening:** the arms that run html now catch the refusal as well as a survivor. A mutant that refused every document used to crash the self-test with a traceback. `mutation-check` scored that as a clean kill with no arm named, because its crash heuristic reads Node error text only.
- **Mutants:** `tools/mutation-envelope.sh --test 'PYTHONDONTWRITEBYTECODE=1 python3 tools/redact-artifact.py --verify'`, the registry row's `verify` command, at `cd934bec3`, all sixteen killed. The failing arms are named from the same command on a patched copy, and no row ends in a traceback.

| mutant (`tools/redact-artifact.py`) | verdict | failing arms |
|---|---|---|
| `dropped = bool(tail) and …` → without `bool(tail)` | KILLED, 13 arms | broad by construction. Without the tail check, any document whose `close()` fires no handler counts as dropped, so every arm expecting a clean html result fails: 11 well-formed/examined documents, the html redaction arm, and the whitespace join. No arm singles this row out |
| `dropped = bool(tail)` (any tail) | KILLED, 6 arms | the six examined tails |
| the refusal disabled (`dropped = False`) | KILLED, 8 arms | the six stall prefixes, the pre-existing malformed arm, and leak-then-stall (its "may be incomplete" note) |
| the old predicate (`not positions(out, kind)`) restored | KILLED, 16 arms | 11 well-formed/examined documents (the nine positionless ones, CDATA, the XML declaration); leak-then-stall; four of the stall prefixes (the doctype prefix no longer yields a position, so the old predicate refuses it too) |
| cross-block hits not collected | KILLED, 6 arms | four `text runs` arms, the label-ordinal and once-per-run arms |
| every occurrence reported as across-block | KILLED, 2 arms | one leak one label; `structure_survivors reports every term…` |
| whitespace-only runs dropped from the join | KILLED, 1 arm | whitespace between blocks |
| the label names the next run | KILLED, 2 arms | the label-ordinal and once-per-run arms |
| the label renamed to `text[` | KILLED, 6 arms | the four inline/table `text runs` arms, label-ordinal, once-per-run |
| no dedupe | KILLED, 1 arm | once per run |
| `unknown_decl` not counted (renamed away) | KILLED, 1 arm | the unclosed CDATA section |
| `if survivors` → `if survivors and not dropped` (round 1, SF-1) | KILLED, 1 arm | leak-then-stall |
| the "may be incomplete" note never added | KILLED, 1 arm | leak-then-stall |
| `handle_decl` renamed away (round 1, SF-2) | KILLED, 1 arm | the unclosed doctype |
| `handle_pi` renamed away (round 1, SF-2) | KILLED, 1 arm | the unclosed XML declaration |
| `dropped = kind in DECODING_TYPES and …` (round 1, SF-3) | KILLED, 1 arm | the mermaid label holding `<` |

The plan file is not committed. Its needles (`OLD::NEW`), none of which ends in `:`, because an OLD ending in `:` meets the `::` separator as `:::` and splits wrong. Four rows first came back `INVALID` for that reason; the parse gate caught it, so nothing passed silently:

```
870-dropped-without-a-tail         self.dropped = bool(tail) and self._events == before::self.dropped = self._events == before
870-dropped-on-any-tail            self.dropped = bool(tail) and self._events == before::self.dropped = bool(tail)
870-refusal-off                    dropped = kind == "html" and _parse_html(out).dropped::dropped = False
870-old-predicate-restored         dropped = kind == "html" and _parse_html(out).dropped::dropped = kind in DECODING_TYPES and out.strip() and not positions(out, kind)
870-across-block-off               hits.extend(_parse_html(text).across_block(terms))::hits.extend([])
870-every-occurrence-across-block  if i + len(t) > offsets[k] + len(self._runs[k][1])::if True
870-whitespace-dropped-from-join   self._runs.append((self._run_start, joined))::self._runs.append((self._run_start, joined.strip() and joined))
870-label-names-next-run           label = f"text-across-block[{self._runs[k][0]}]::label = f"text-across-block[{self._runs[k + 1][0]}]
870-label-renamed-text             label = f"text-across-block[::label = f"text[
870-no-dedupe                      if label not in hits::if True
870-cdata-not-counted                  def unknown_decl(self, data::    def _unused_unknown_decl(self, data
909-survivors-yield-to-drop            if survivors::    if survivors and not dropped
909-incomplete-note-dropped        if dropped else ""::if False else ""
909-decl-not-counted                   def handle_decl(self, decl::    def _unused_handle_decl(self, decl
909-pi-not-counted                     def handle_pi(self, data::    def _unused_handle_pi(self, data
909-refusal-widened                dropped = kind == "html" and::dropped = kind in DECODING_TYPES and
```

## Not changed, and where it went

- **The event counters in `handle_starttag` / `handle_endtag`** are not separately pinned. `close()` counting an emitted tag as "examined" is the right semantics, but no measured shape makes `close()` emit a bare tag. The line `self._events += 1` occurs seven times. The decl, PI and CDATA counters are pinned by renaming their handler away. No mutant isolates the tag counters: renaming `handle_starttag` away also removes attribute positions and run flushing.
- **A `text-across-block[N]` label can name a whitespace-only run.** It happens only for a term with leading whitespace (round 1, N-5), and it is cosmetic.
- **Two pre-existing misses of the same "what markup can decide" class**, identical on `main`: CDATA content inside SVG never joins a run, and `script`/`style` content does. Filed as #910.
- **`skills/redact-visualization-for-disclosure/SKILL.md`** documents this tool's labels and exit-2 causes, and is tracked by #901. After this PR it should list `text-across-block[N]` beside `text[N]`, and the refusal as "input the parser dropped". #901 is the place for that, not this PR.
- **Exit codes.** The contract stays 0 clean / 1 a term survived / 2 could not measure. `text-across-block` is an exit 1.

## Review

This is code in `tools/`, where the non-required `tools-verify` context runs its self-test, so it gets an advocatus round. Before the push, the commit message and this body go to the support session's claim check.

**Claim check before the push**, on the first local draft: `BLOCKING 0 · SHOULD-FIX 2 · NOTE 1`. The mutant table reproduced exactly, and the before/after, the decl/pi argument and the `_pending` paragraph were re-derived.
- **SF-1:** CDATA was not a drop. `close()` emits an unclosed section to `unknown_decl`, which `_Positions` did not implement, so the tool refused a raw substring. It was found with a parser implementing every handler, and I re-derived it: counting `unknown_decl` makes the new CDATA arm pass, and the rename-away mutant fails exactly that arm. Fixed before the push, with the table row and the comment restated.
- **SF-2:** this body said `tools-verify` has no `setup-python` step. It has one. The sentence now names the step and what was read from source.
- **NOTE-1:** row 1's "discriminating arms" wording was wrong. With the tail check gone, every html arm fails and none singles the row out.

**Claim check, second pass:** `BLOCKING 0 · SHOULD-FIX 0 · NOTE 1`, clean. It re-derived every changed clause and all eleven rows. It also read upstream `parser.py` at every tag from v3.12.3 to v3.12.14 (0 matches for `_pending` until v3.12.14, then 17), and confirmed that counting `unknown_decl` lets no real drop through. The NOTE was three unwrapped lines, now rewrapped. The table was re-measured after that, at `cfcea15dc`.

**Advocatus round 1 at `cfcea15dc`:** `BLOCKING=0 SHOULD-FIX=3 NOTE=5`, mergeable. It found that no malformed document passes and no well-formed one refuses (42 shapes), no leak `main` caught is missed (86-row corpus), and all labels are stable across decl/PI/CDATA. All eleven rows reproduced. Fixed in `cd934bec3`:
- **SF-1:** the drop check ran before survivors, so a known leak followed by a stall went from `main`'s exit 1 to exit 2, and a caller branching on 1 heard nothing. Re-derived first: the new arm failed on the old order. Survivors come first now, with a "may be incomplete" note. Leak-then-stall exits 1 on this branch and on `main`.
- **SF-2:** the decl and PI counters were unpinned, and this body's table claimed them. Two arms now pin them; both passed before the fix, so they pin existing behaviour.
- **SF-3:** "the one type that can reach it" had no arm. A mermaid label holding `<` now pins it.
- **N-1:** seven counters, not six (corrected above). **N-2:** "did not parse" was #869's change (corrected). **N-3:** the adjacent pre-existing class is filed as #910. **N-4:** `_parse_html`'s docstring now says why an instance is fed once. **N-5:** cosmetic, noted under Not changed.

**Claim check of the round-1 fixes**, before their push: `BLOCKING 0 · SHOULD-FIX 0 · NOTE 2`, clean. It re-derived every changed clause and all sixteen rows, and it read the CI run's Python from its log. The two notes are applied above: the `r1` cells now show the "may be incomplete" note, and the tag-counter sentence is tighter.


pjt222:
**Round 2 (advocatus confirm) at `cd934bec3b9f9e8be506cb170306b4408b673590`: `BLOCKING=0 SHOULD-FIX=0 NOTE=1` — mergeable.**

Recorded the next day: the report arrived at 2026-09-23T21:09Z, after the session had last updated this PR, and was recovered verbatim from the reviewer transcript with `tools/agent-report.mjs --nth 2`.

<details><summary>Full round-2 report</summary>

GATE: BLOCKING=0 SHOULD-FIX=0 NOTE=1 — **confirm, mergeable.**

## Provenance

Bundle `BUNDLE_SHA` = `cd934bec3b9f9e8be506cb170306b4408b673590`, status `(no output)`; worktree HEAD and PR head are the same sha; `origin/main..HEAD` is exactly the two commits. `branch2/` in the lab was extracted with `git show cd934bec3:tools/…`, hash `47d947668` = `cd934bec3:tools/redact-artifact.py`, lib unchanged (`c6328fe7a`). Self-test on the copy: `79/79 arm(s) passed`, arm list counts 79. CI at the head: all five required contexts and `tools-verify` **pass**; the run's log says `Successfully set up CPython (3.12.14)` at `headSha cd934bec3`. Nothing written outside `scratchpad/advocatus-909/`; no `__pycache__`.

## (a) SF-1 shapes and the ordering mutant

Through the CLI (`--type html --mapping {acme_secret:…} --also-deny acme_secret`), main / r1 (`cfcea15dc`) / r2 (`cd934bec3`):

```
known leak + malformed tail       main: 1 FAILED text[1]:acme_secret, text[1]:acme_secret
                                  r1:   2 REFUSED --type html: the parser dropped input…
                                  r2:   1 FAILED text[1]:acme_secret, text[1]:acme_secret; the parser also dropped input it could not tokenise, so this list may be incomplete
known leak + unclosed start tag   main 1 / r1 2 / r2 1, same message shape
```

Both SF-1 shapes are back at exit 1 with the position named and the incompleteness note appended. Two controls behave as the reorder predicts: a **raw** leak in a stalled document (`<p>acme_secret</p>` + stall) exits 1 via `whole-text` with no note on r2, which is correct — the whole-text tier scans the raw tail too, so its list is complete; only the structure tier's can be partial. And a clean document with a stall stays `2 REFUSED` on r2.

Ordering mutant: `mine-r2-old-order-restored` (the `dropped` refusal moved back ahead of `if survivors`) is KILLED — by the leak-then-stall arm, plus a second arm ("raises UnmeasurableError, NOT RedactionError") that fired only because my mutant's message lacked `COULD-NOT-MEASURE`; that is my instrument, not a second discriminator. The PR's own row 12 (`if survivors and not dropped`) is the cleaner form and dies to exactly the one arm. Round 1's `mine-ordering-survivors-first` no longer applies (its needle is the new code).

## (b) Does the reorder move anything else?

Round-1 86-row corpus, r1 vs r2 (`exp6_r2.py`): **84 of 86 identical; the two diffs are exactly the SF-1 rows.** main vs r2: 51 of 86 identical (same count as main vs r1) and no row where main FAILED and r2 did not. md/text/mermaid rows byte-identical across all three.

One extra shape outside the corpus, reported as confirmation rather than a finding: across-block leak followed by a stall (`<div>acme_</div><div>secret</div>` + stall) goes main 0 / r1 2 / r2 `1 FAILED text-across-block[1]:acme_secret; the parser also dropped…`. Same class as SF-1; the body's rule sentence ("a leak the tier finds is exit 1 even when the parser also dropped input after it") covers it, its table just does not list it.

## (c) The body's round-1 sentences against my report

Every sentence checks out against what I measured:

- **Mutants**: all 16 rows reproduce exactly on `branch2/` copies — 13/6/8/16/6/2/1/2/6/1/1 and 1/1/1/1/1 — no traceback. Row 1's drop from 27 to 13 is what the reorder predicts: docs with a survivor now exit 1 before the refusal, so only the 11 well-formed/examined docs, the html redaction arm and the whitespace join remain. My round-1 survivors that motivated SF-2/SF-3 now die to the new arms: counter deleted in `handle_decl` → 1 arm (unclosed doctype), in `handle_pi` → 1 arm (unclosed XML declaration), predicate widened → 1 arm (mermaid `a["a<b"]`).
- **79/79 (46 on main)**, seven `_events += 1` sites, "did not parse" attributed to #869, `r1` cells carrying the note, before/after "after is `cd934bec3`" — all match.
- **#910** exists, OPEN, names exactly the two shapes I measured (SVG CDATA-split and script/style-split), says they are identical on main, and says the filing session did not re-derive them — honest.
- **`_parse_html` docstring** matches the N-4 measurement (second `feed()` sits in `_pending`, merged after `_Positions.close` read `rawdata`).
- **"both passed before the fix, so they pin existing behaviour"** (SF-2 arms) — consistent with round 1, where r1 was clean on both documents.
- **CI**: the body cites the 3.12.14 log line at `cfcea15dc`; the `cd934bec3` run says the same.

## NOTE

**N-1 (wording, "Not changed" paragraph):** "No mutant isolates the tag counters: renaming `handle_starttag` away also removes attribute positions and run flushing." That conflates *cannot be targeted* with *cannot be killed*. A mutant deleting only the counter line does isolate each one — `mine-r2-events-not-counted-on-starttag` / `-endtag` — and both **SURVIVED 79/79**, which is the equivalent-mutant outcome the body's own argument predicts (no measured shape makes `close()` emit a bare tag; the end-mode branch at `parser.py:274-304` never calls a tag handler). Under this repository's mutant discipline an equivalent survivor is named as such. Proposed: "a mutant deleting only the counter line survives, and is equivalent: no measured shape makes `close()` emit a bare tag." Does not block.

Not filed: a rewording of the "may be incomplete" note survived, but it kept the word `dropped` that the arm asserts, so it is equivalent for the property; the PR's row 13 removes the note entirely and dies.

Lab: `/tmp/claude-1000/-mnt-d-dev-p-agent-almanac/49a9d91f-a338-44ce-8046-90b72123b8e2/scratchpad/advocatus-909/lab.HKu49L/{exp6_r2.py,mutate2.py,plan_r2_mine.tsv,branch2/}`.

GATE: BLOCKING=0 SHOULD-FIX=0 NOTE=1 — confirm, mergeable.
</details>

